PR #11, opened 2018-08-03
Add props to set custom min and max zoom levels

MattyBalaam:
This pull request brings in customisable minimum and maximum scrollwheel zoom values.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -164,7 +164,7 @@ class Cropper extends React.Component {
   }
 
   setNewZoom = zoom => {
-    const newZoom = Math.min(MAX_ZOOM, Math.max(zoom, MIN_ZOOM))
+    const newZoom = Math.min(this.props.minZoom, Math.max(zoom, this.props.maxZoom))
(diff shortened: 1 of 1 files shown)


src/index.js
```diff
@@ -164,7 +164,7 @@ class Cropper extends React.Component {
   }
 
   setNewZoom = zoom => {
-    const newZoom = Math.min(MAX_ZOOM, Math.max(zoom, MIN_ZOOM))
+    const newZoom = Math.min(this.props.minZoom, Math.max(zoom, this.props.maxZoom))
```
ValentinH: you switched min and max here ;)

ValentinH: Thanks to the e2e tests to detect this. 🙂

ValentinH:
Nice, I was sure that somebody was going to contribute this 😆.

src/index.js
MattyBalaam: Ah, sorry, will recommit

ValentinH:
Thanks!

PR #14, opened 2018-09-17
Add cropType property that allows round cropper

claudiuandrei:
Add a new property to the Cropper component, **cropType**. The new property will allow to pass in different cropper styles: _round_, _rect_ for the current change.

My use case is that I need a round cropper to replace [Croppie](https://foliotek.github.io/Croppie/), react-easy-crop is perfect and really well implemented but it is missing this. Let me know if this is a feature you want to add and if yes if this is the interface you are looking for or you want some other changes.

### DIFF
--- a/src/styles.js
+++ b/src/styles.js
@@ -32,8 +32,7 @@ const cropperLines = {
   position: 'absolute',
   border: lineBorder,
 }
-
-export const CropArea = styled('div')({
+const cropperArea = {
   position: 'absolute',
   left: '50%',
   top: '50%',
@@ -41,6 +40,10 @@ export const CropArea = styled('div')({
   border: lineBorder,
   boxSizing: 'border-box',
   boxShadow: '0 0 0 9999em rgba(0, 0, 0, 0.5)',
+}
+
+export const RectCropArea = styled('div')({
+  ...cropperArea,
   '&::before': {
     ...cropperLines,
     top: 0,
@@ -59,4 +62,9 @@ export const CropArea = styled('div')({
     borderLeft: 0,
     borderRight: 0,
   },
+})
+
+export const RoundCropArea = styled('div')({
+  ...cropperArea,
+  borderRadius: '50%',
 })
--- a/src/styles.js
+++ b/src/styles.js
@@ -66,16 +66,15 @@ const roundShape = {
   borderRadius: '50%',
 }
 
-export const createCropArea = ({ shape, showGrid }) =>
-  styled('div')({
-    ...(() => {
-      switch (shape) {
-        case 'round':
-          return { ...cropperArea, ...roundShape }
-        case 'rect':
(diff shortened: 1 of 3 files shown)


ValentinH:
Hey!

Thanks for contributing and for the nice comment 👍 

Here's a sandbox including your changes: https://codesandbox.io/s/7zw1ow23yj

I guess your main use case is for round images, how do you expect the lib to behave for rectangle images (like in the sandbox above)?

Regarding the API, this is fine to me. Maybe we could also add a toggle to display the "third-lines"?

claudiuandrei:
Hi @ValentinH 

Thanks for setting up a sandbox.

I think the library behaves as it should for rectangle images, the only thing is that there was no visual indication for the round images.

For **onCropComplete** nothing needs to change, as the cropper will just return the coordinates and the parent component will decide how to crop based on its own state (round or rectangle). I actually like that there is no assumption on that in the cropper. Basically **cropType** is about the visual indication (maybe I can name it differently like **cropArea** or **cropAreaShape**).

The "third-lines" are actually a cool feature, it might me great if they are behind a prop, i will try to do this. I tried to keep them with the round cropper but it looks funky because they get out of the area and I have to idea how to limit them to the circle. If you know how to do that would be awesome. I will research too.

ValentinH:
Sorry I was wrong about rectangles images having ellipses instead of circle: I needed to set the `aspect` prop to 1.

Regarding the props API, I'm not a big fan of your latest suggestions to be honest. I would keep your first idea of `cropType` (or `cropShape` maybe) and simply add a `showGrid` property as well.

Regarding `onCropComplete`'s API, if it works for you then it's fine to me 🙂

claudiuandrei:
@ValentinH Done! Thanks for the suggestion.

src/styles.js
```diff
@@ -59,4 +61,21 @@ export const CropArea = styled('div')({
     borderLeft: 0,
     borderRight: 0,
   },
-})
+}
+const roundShape = {
+  borderRadius: '50%',
+}
+
+export const createCropArea = ({ shape, showGrid }) =>
```
ValentinH: You can actually replace this function by a styled component using emotion props api: https://emotion.sh/docs/styled#object-styles

Like this, you would then only have to pass the props to `CropArea`:
```jsx
<CropArea
  cropShape={cropShape}
  showGrid={showGrid}
  style={{
    width: this.state.cropSize.width,
    height: this.state.cropSize.height,
  }}
  data-testid="cropper"
  />
```

and it would avoid recreating the whole emotion component at each render.

claudiuandrei:
@ValentinH Thanks, that was a great suggestion, I was not aware of that possibility, much cleaner.

claudiuandrei:
Thank you for the help and responsiveness.

PR #16, opened 2018-10-20
Add styling props

tafelito:
This PR adds styling props to the container, cropping area and image

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -270,7 +270,7 @@ class Cropper extends React.Component {
             transform: `translate(${x}px, ${y}px) scale(${zoom})`,
           }}
           imageStyle={imageStyle}
-          className={cropAreaClassName}
+          className={imageClassName}
         />
         {this.state.cropSize && (
           <CropArea
--- a/src/index.js
+++ b/src/index.js
@@ -248,6 +248,7 @@ class Cropper extends React.Component {
       cropShape,
       showGrid,
       style: { containerStyle, cropAreaStyle, imageStyle },
+      classes: { containerClassName, cropAreaClassName, imageClassName },
     } = this.props
 
     return (
@@ -258,6 +259,7 @@ class Cropper extends React.Component {
         innerRef={el => (this.container = el)}
         data-testid="container"
         containerStyle={containerStyle}
+        className={containerClassName}
       >
         <Img
           src={this.props.image}
@@ -268,6 +270,7 @@ class Cropper extends React.Component {
             transform: `translate(${x}px, ${y}px) scale(${zoom})`,
           }}
           imageStyle={imageStyle}
+          className={cropAreaClassName}
         />
         {this.state.cropSize && (
           <CropArea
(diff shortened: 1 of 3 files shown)


ValentinH:
In order to use those props, you need to pass a style object right? What I had in mind was more something like `containerClass`, `cropAreaClass`, etc... Or a `classes` property that would take an object like this:
```
{
  container: 'xyz',
  cropArea: 'abc'
}
```

This would be simpler to integrate with any styling solution, CSS-in-JSS or not.

ValentinH:
In order to use those props, you need to pass a style object right? What I had in mind was more something like `containerClass`, `cropAreaClass`, etc... Or a `classes` property that would take an object like this:
```
{
  container: 'xyz',
  cropArea: 'abc'
}
```

This would be simpler to integrate with any styling solution, CSS-in-JSS or not.

tafelito:
I don't use classes. In fact I'm using react-native-web so I wouldn't be able to pass a class or a className. But either way, that could be added as well so you could either pass a class or a style and it will work in both cases

What do you think?

ValentinH:
I understand. On the same idea as the one I suggested with classes, could you group all of them under a `styles` or `customStyle` property that would take an key/value map of styles?

Also could you please update the ReadMe so other people know how to use this new feature 🙂

tafelito:
I haven't really used emotion or styled-components so what would be the best way to merge the classes with the default styles in this case?

src/index.js
```diff
@@ -268,6 +270,7 @@ class Cropper extends React.Component {
             transform: `translate(${x}px, ${y}px) scale(${zoom})`,
           }}
           imageStyle={imageStyle}
+          className={cropAreaClassName}
```
ValentinH: I think it should be `imageClassName` here 😄

src/index.js
```diff
@@ -279,6 +282,7 @@ class Cropper extends React.Component {
             }}
             data-testid="cropper"
             cropAreaStyle={cropAreaStyle}
+            className={imageClassName}
```
ValentinH: Opposite here: should be `cropAreaClassName`

src/index.js
```diff
@@ -265,6 +269,8 @@ class Cropper extends React.Component {
           style={{
             transform: `translate(${x}px, ${y}px) scale(${zoom})`,
           }}
+          imageStyle={imageStyle}
+          className={cropAreaClassName}
```
ValentinH:
```suggestion
          className={imageClassName}
```

src/index.js
```diff
@@ -275,6 +281,8 @@ class Cropper extends React.Component {
               height: this.state.cropSize.height,
             }}
             data-testid="cropper"
+            cropAreaStyle={cropAreaStyle}
+            className={imageClassName}
```
ValentinH:
```suggestion
            className={cropAreaClassName}
```

ValentinH:
Thanks for this PR and the adjustements while I was a bit slow to answer 👍

tafelito:
close #16

tafelito:
@ValentinH, when do you think you'll publish a new version?

ValentinH:

PR #17, opened 2018-10-20
Invert props order check on update

tafelito:
This fixes an issue that prevents updating at the same time zoom and aspect ratio. Because `computeSizes` also recomputes crop position, if aspect ratio and/or zoom are changed, both values will be updated


ValentinH:
Thanks for this :)

ValentinH:

PR #18, opened 2018-10-23
add onError prop to Img

bexoss:
add onError prop to Img

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -261,6 +261,7 @@ class Cropper extends React.Component {
           src={this.props.image}
           innerRef={el => (this.image = el)}
           onLoad={this.onImgLoad}
+          onError={this.onImgError}
           alt=""
           style={{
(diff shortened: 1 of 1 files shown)


bexoss:
@ValentinH Oh, I forgot `props.`, Could you add the onError handler for ``?  
I want to use fallback image if the src not exist.

ValentinH:
This looks good to me, just before merging, could you add this to the documentation?

bexoss:
Okay, I added pull request. Should I used same branch? Actually I don't have enough experiences to pull request to other project. So it little confused me. Anyway I added pull request if you find any wrong please edit it correctly :) I want to push commit to this same branch but I found the documentation updated few hours ago so I created new one to new docs. I don't know git handle it automatically or not :(

ValentinH:
Indeed, you should have done it in the same PR. But let's merge both :)

ValentinH:

PR #23, opened 2019-02-26
Add a crossOrigin pass-through property

nbibler:
The usage of [crossOrigin](https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/crossOrigin) on image tags allows Chrome and Safari to optionally send Origin, Origin + Authentication, or no information (the default) to third-party sites during image loading. This is required when enforcing CORS.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,13 +1,12 @@
-import { Container, CropArea, Img } from './styles'
+import React from 'react'
 import {
+  getCropSize,
+  restrictPosition,
+  getDistanceBetweenPoints,
   computeCroppedArea,
   getCenter,
-  getCropSize,
-  getDistanceBetweenPoints,
-  restrictPosition,
 } from './helpers'
-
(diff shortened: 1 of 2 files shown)


src/index.js
```diff
@@ -1,12 +1,13 @@
-import React from 'react'
+import { Container, CropArea, Img } from './styles'
```
ValentinH: I'm not a big fan of this import reordering. I prefer to have the more global ones first and then the local imports.

nbibler: Ahh. No problem. My editor automatically sorted them when I saved the file. I'll undo it and push a new commit in a few.

nbibler:
Import orders were reset. Feel free to squash into a single commit if you like. :)

ValentinH:
Released as 1.8.0

nbibler:
Thank you!

PR #24, opened 2019-03-08
Disable page scrolling on iOS when user scrolls inside of crop component

yokelpole:
Currently on iOS the entire page scrolls when a user attempts to move the cropped area:
before

This makes it so that we prevent that behaviour (there is a little bit of scrolling on the page itself to show that functionality is unaffected by my change):
fixed

(Sorry for the choppiness in the examples - the iOS simulator is a little slow. I have verified that this happens on an iPhone 7 with iOS 12.)


ValentinH:
Hey thanks for this. 

Indeed, I might have never tried this on a scrollable page (I was always in a fullscreens pop-up).

Before merging, I'll try it on other devices just to be sure nothing else is affected. 🙂

ValentinH:
Tested and merged! 👍 

BTW your app looks great. If it's public, I'd love to see how you use this library 😃

ValentinH:
Published as 1.8.1

yokelpole:
Sure thing! We're using the library at http://www.change.org for our smart images uploader. It's a feature that allows users to upload multiple pictures to their petitions so that they can test what image is the most effective for link sharing.

There's no easily available way to access the page without starting a petition, but if you do start a petition you can access the page by going to `/p/{your-petition-slug}/smart` and you'll be able to see it in action there.

ValentinH:
This makes me so happy to see this project for such a cool website! 🎉

ValentinH:
BTW, the french translation of the title is incorrect:
image

It should be "Testez d'autres images..."

PR #32, opened 2019-04-24
Add restrictPosition prop

### LINKED ISSUE #31: Moving images with zoom smaller than 1 not possible
## Are you reporting a bug?

- Please create a CodeSandbox that demonstrates your issue. You can fork the [Basic example](https://codesandbox.io/s/q80jom5ql6).

  https://codesandbox.io/s/92r5r56zqr?fontsize=14

- Provide the steps to reproduce the issue, e.g.:

  1.  Try to move the image

  **Observed behaviour:** It is not possible to move the image within the cropping boundaries if the zoom is smaller than `1`.

  **Expected behaviour:** Zooms smaller than `1` should be supported in order to not only crop, but also change the aspect ratio of an image. In my application I'd like to get images with a constant aspect ratio. For the best UX I'd like the user to be able to also just use any image with a white background. Vertical images are especially a problem. A possible implementation could be to modify the `restrictPosition` function. I do a lot of JS myself, but if you have a quick fix for this it would be great to see this improvement to this overall very nice project!

nicklasfrahm:
This PR fixes #31. It adds a new `restrictPosition` prop that allows the user to determine if the position should be restricted or not. Additionally the `README.md` is updated to document this new prop. Minor fixes to the prop table have been applied to align the casing.


ValentinH:
Thanks for this!

ValentinH:
Released at v1.11.0: https://github.com/ricardo-ch/react-easy-crop/releases/tag/v1.11.0 🎉

nicklasfrahm:
You are welcome! I am just getting started to open-source contributions, this was a great opportunity, because your contribution guidelines are very clear, you respond very fast and the fix is not super complex.

ValentinH:
I'm happy that it felt easy for you! 🙂 

Eve, if it is an easy fix, it's always good to contribute and to get started with open-source! 🎉

nicklasfrahm:
Very much so! I'd love to work full-time in open-source at some point if it is financially sustainable, because I love the concept. Everybody can fix stuff and read everybody's code. 🚀 

Need to work on my skillset though. As a Mechatronics engineer fullstack JavaScript is a skill that gives you a lot of funny looks from colleagues and fellow students.

ValentinH:
A really important thing that you did pretty well here was to clearly communicate your issue, this is really appreciated from maintainers 🙂

nicklasfrahm:
Thank you. Totally agree. This is also something that is not always very easy. I have noticed it at work quite a few times. Somebody writes an issue and when you pick it up as a developer it can be quite strange if you want to solve this bug, but you don't even know what it's about. And this just happens, because the issue sometimes seems so clear to you, that you do not think about the 10 assumptions that you made along the way.

PR #42, opened 2019-06-07
added onInteractionBegan callback to onWheel and onDrag

Alba-C:
While onInteractionEnded() may be more useful in the long run, onInteractionBegan() seems to accomplish our immediate task and looks like an easier solution to implement.


src/index.js
```diff
@@ -171,6 +171,7 @@ class Cropper extends React.Component {
         ? restrictPosition(requestedPosition, this.imageSize, this.state.cropSize, this.props.zoom)
         : requestedPosition
       this.props.onCropChange(newPosition)
```
ValentinH: Why not calling it inside `onDragStart`?

src/index.js
```diff
@@ -24,6 +24,7 @@ class Cropper extends React.Component {
   rafZoomTimeout = null
   state = {
     cropSize: null,
+    onWheelStart: false,
```
ValentinH: I would name it something like `hasWheelJust Started` because current naming made me think it was a callback. 😀

src/index.js
```diff
@@ -206,6 +209,15 @@ class Cropper extends React.Component {
     const point = Cropper.getMousePoint(e)
     const newZoom = this.props.zoom - (e.deltaY * this.props.zoomSpeed) / 200
     this.setNewZoom(newZoom, point)
+
+    !this.state.onWheelStart &&
```
ValentinH: Could you please use an if block here for better readability?

src/index.js
```diff
@@ -206,6 +209,15 @@ class Cropper extends React.Component {
...
+      this.setState({ onWheelStart: true }, () => this.props.onInteractionStart())
+
+    clearTimeout(this.wheelTimer)
```
ValentinH: Please also clear the timeout in `componentWillUnmount()`.

src/index.js
```diff
@@ -206,6 +209,15 @@ class Cropper extends React.Component {
...
+    this.wheelTimer = setTimeout(
+      () => this.setState({ onWheelStart: false }, () => this.props.onInteractionEnd()),
+      250
```
ValentinH: Is this enough? I still have to play with it to forge an opinion, I'm not on the computer now ^^

Alba-C:
I took your advice and moved the first onInteractionStart() call from onDrag to onDragStart and didn't need onPinchStart, because it also calls onDragStart.  Inside onWheel I added a state check for onInteractionStart so it doesn't call it too many times and paired that with a timer for onInteractionEnd so it only fires after the user has stopped scrolling the wheel for 250ms.

ValentinH:
I've just run the code and 250ms looks good to me too.

One more thing, please remove the `package.lock` file, this project uses Yarn. (You can add it to the .gitignore if you want :) )

Alba-C:
All changes complete, let me know if you see anything else.

ValentinH:
@all-contributors please add @Alba-C for code 🎉

PR #45, opened 2019-07-04
Do not compute zoom if picture is still loading

ersefuril:
Hey there !

Zooming on a picture that is not loaded yet is throwing ```helpers.js Uncaught TypeError: Cannot read property 'width' of null```.

So I did a quick fix to solve that issue !

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,6 @@
     "cy:run": "cypress run",
     "cy:ci": "cypress run --record",
     "prepublishOnly": "yarn build",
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -77,5 +76,10 @@
       "prettier --write",
       "git add"
     ]
+  },
```
ValentinH: Do you need to make this change?

ersefuril:
My bad, I'll create a new PR without all of these changes ! 
Thanks :)

PR #51, opened 2019-08-19
Add rotation crop preview

MattyBalaam:
This adds an example of how to use Canvas to generate an image based on combined rotatation and crop values.

### DIFF
--- a/docs/src/components/Demo/index.js
+++ b/docs/src/components/Demo/index.js
@@ -65,8 +65,8 @@ const Demo = ({ classes }) => {
   }, [])
 
   const showCroppedImage = useCallback(async () => {
-    const croppedImage = await getCroppedImg(dogImg, croppedAreaPixels)
-    setCroppedImage(croppedImage, rotation)
+    const croppedImage = await getCroppedImg(dogImg, croppedAreaPixels, rotation)
+    setCroppedImage(croppedImage)
   }, [croppedAreaPixels])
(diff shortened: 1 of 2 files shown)


MattyBalaam:
OK, I’ve run the demo now, and all seems to work as expected.

docs/src/components/Demo/index.js
```diff
@@ -65,7 +65,7 @@ const Demo = ({ classes }) => {
   }, [])
 
   const showCroppedImage = useCallback(async () => {
-    const croppedImage = await getCroppedImg(dogImg, croppedAreaPixels)
+    const croppedImage = await getCroppedImg(dogImg, croppedAreaPixels, rotation)
     setCroppedImage(croppedImage)
   }, [croppedAreaPixels])
```
ValentinH: You need to add the `rotation` to the deps array of `useCallback()`. Otherwise, if you change the rotation twice in a row, the output wont't have the right rotation.

MattyBalaam: Good spot, I’m so used to have the`rule-of-hooks` linter autofix that for me I never think to check.

ValentinH: Yes, same here. 😄 We need to add this to this project.

ValentinH:
Thanks a lot! I'm deploying this to the documentation website :)

Issue #52, opened 2019-08-19
Improve edge detection when rotation is used


ValentinH:
The current `restrictPosition()` function ([here](https://github.com/ricardo-ch/react-easy-crop/blob/master/src/helpers.js#L28)) does not take care of the `rotation` prop and is too restrictive.

We need to handle the rotation properly to have correct edge detections.

MattyBalaam:
Not PR ready, but I had a play around this. It may or not be useful. I have the crop-area successfully tracking-around the area an image has roatated in. I’m _guessing_ that instead of cropping like this, the ideal would be to crop without showing any areas outside on angles outside of 0/90/180/270

https://github.com/MattyBalaam/react-easy-crop/commit/f2bdb2d1c39e22ab5090182f1540847d5b992358

ValentinH:
Thanks for this, I'll check it later today 🙂

ValentinH:
I just checked it and I'm actually not sure to understand why you are modifying the crop area size based on the rotation. This behaviour does not feel really familiar to me. 🤔 

I would expect to keep the crop area fixed but only adapt the rules of the edge detection.

Anyway, thanks for giving this a try! 👍

MattyBalaam:
I wasn't sure where to start as I'm not totally sure how everything works conceptually behind the scenes, so just tried at the first place I found I could manipulate things. I'll take a look at the edge detection tomorrow.

MattyBalaam:
I adapted the code to be in the restrict function based on the rule of keeping at least part of the image without the crop area.

https://github.com/MattyBalaam/react-easy-crop/commit/0e665daa017a45628069f7994f684c2917a7aeb1

A caveat with this is that when the demo portrait image is rotated 90 degrees then it is possible to manipulate the zoom so that the width of the image becomes smaller than the crop. Do you have pointers about the best place to adjust the zoom settings in this situation?

![Screen Shot 2019-08-22 at 17 10 52](https://user-images.githubusercontent.com/1246923/63531162-d4f45e00-c4ff-11e9-9ba6-0da9d00faf68.png)

I also spent quite a while attempting to figure out way to restrict based on keeping the crop entirely within the image, but I’ve not been able to quite work out how to convert the co-ordimates properly. I might take another look, but it has been quite a slog trying to work it out.

MattyBalaam:
This modification keeps the crop within the image - however the crop itself changes size when the rotation is changed.

```
[...]

PR #53, opened 2019-08-22
Constrain image when rotated

MattyBalaam:
Adds support for constraining an image in the crop area when rotated.

See https://github.com/ricardo-ch/react-easy-crop/issues/52

### DIFF
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -25,10 +25,12 @@ export function getCropSize(imgWidth, imgHeight, aspect) {
  * @param {number} zoom zoom value
  * @returns {{x: number, y number}}
  */
-export function restrictPosition(position, imageSize, cropSize, zoom) {
+export function restrictPosition(position, imageSize, cropSize, zoom, rotation) {
+  const { width, height } = translateSize(imageSize.width, imageSize.height, rotation)
+
   return {
-    x: restrictPositionCoord(position.x, imageSize.width, cropSize.width, zoom),
-    y: restrictPositionCoord(position.y, imageSize.height, cropSize.height, zoom),
+    x: restrictPositionCoord(position.x, width, cropSize.width, zoom),
+    y: restrictPositionCoord(position.y, height, cropSize.height, zoom),
   }
 }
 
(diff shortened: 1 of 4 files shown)


ValentinH:
I've played again with your demo and I actually think we should combine this one and the one you made before where the crop size was changed. 

I'm not sure where to draw the limit but I think the crop size should only be modified if the image width (or height) gets smaller than the cropper. Another limit that might be easier to define would be to prevent the crop size being bigger than the container ones. What do you think?

For example, this is what should the crop size be when rotation is 0:
image

And this is for 90deg rotation:
image


If you want there is also a vertical image (cat.jpeg) in the example folder 😉

MattyBalaam:
I have managed to accomplish this (at least on landscape, I’ll take a look at portrait later when I have some time) by adding an additional check to the `getCropSize` function.

ValentinH:
I don't know if you pushed your latest changes, but I actually see the opposite behavior when the image is close to 90deg: it crop size gets even wider.
image

MattyBalaam:
Apologies, try again now.

MattyBalaam:
I’ve looked at the cat image now and I’ll need to apply some extra logic to handle the portrait case.

ValentinH:
Yes this looks great now! Just one question, do we want to also resize the crop area if the image is zoomed? It's actually not needed but might be simpler: otherwise, if the user zooms, rotate, and zooms back, we need to resize again while zooming back (don't know if this is clear ^^)

MattyBalaam:
I see, so this could be done with an additional check in the crop size calculation which accepts the zoom value perhaps.

MattyBalaam:
I’ve pushed a change purely for handling the sizing on landscape images. I couldn’t get a nice result zooming the crop area.

src/helpers.js
```diff
@@ -25,10 +42,12 @@ export function getCropSize(imgWidth, imgHeight, aspect) {
  * @param {number} zoom zoom value
  * @returns {{x: number, y number}}
  */
-export function restrictPosition(position, imageSize, cropSize, zoom) {
+export function restrictPosition(position, imageSize, cropSize, zoom, rotation) {
```
ValentinH: Here you should set the default value for the rotation to 0.

src/helpers.js
```diff
@@ -157,3 +176,39 @@ export function getCenter(a, b) {
     y: (b.y + a.y) / 2,
   }
 }
+
+function rotate(x, y, xm, ym, a) {
```
ValentinH: Could you use more verbose variable names so that it's easier to understand what each of them do? 🙂

src/helpers.js
```diff
@@ -157,3 +176,39 @@ export function getCenter(a, b) {
...
+  var cos = Math.cos,
```
ValentinH: Also please use `const` here and create a new variable for `a` (rotationInRadians maybe) to avoid overwriting to function arguments.

MattyBalaam: Sure, I was being lazy and had just copied in something I found online, will rewrite to be more descriptive in a couple of days

ValentinH:
BTW I've just added support for rotation gestures on mobile. It should not have any impact with your PR but let's see.

ValentinH:
Hey @mbalaam, just wanted to let you know that I'll be offline for the next 3 weeks starting from tomorrow evening. Thus, if you need this feature soon, we need to wrap this up before :)

MattyBalaam:
Thanks for letting me know, hoping to get some time tomorrow morning on this

MattyBalaam:
I have updated the docs and unit tests. I’ve not used Cypress and won’t have time to learn it in the next day or so unfortunately.

One thing I have noticed is that there will be some floating numbers with a few decimal places. I wondered if it might be a good idea to round these in any helpers that return to a component.

ValentinH:
Wonderful! Thank you so much for taking care of this! ❤️

MattyBalaam:
My pleasure.

ValentinH:
Published with under v1.15: https://github.com/ricardo-ch/react-easy-crop/releases/tag/v1.15.0

PR #89, opened 2020-01-06
fix README props table

simrobin:
Hello, 

Since I'm currently using your library, I was reading the Props section of the README and I saw that the preview was broken because of a blank space.

Thanks for the good work in `2.0.0` 🙂 !


ValentinH:
Damn, I didn't even see this ^^ Thanks for the quick fix!

PR #100, opened 2020-02-20
Add zoomWithScroll prop for prevent scrolling & update README

### LINKED ISSUE #99: Prevent zoom by scroll
Hi,
I would like to prevent changing the zoom by scrolling because in my case it's disturbing the page scrolling.
Does there is support for such an option or do you any idea for me to implement it?
Thanks
BTW, I really like this package 😍 .

Michal-Sh:
Adding the prop `zoomWithScroll` to enable preventing zoom by scroll.
Closes #99


ValentinH:
Thank you really much for this contribution!

This was released with v2.1.0 🎉

Michal-Sh:
It was my pleasure 😄

ValentinH:
@allcontributors please add @Michal-Sh for code

PR #109, opened 2020-03-31
add cropSize change

FillPower1:
I added a couple of lines of code, like nothing broke and works)


ValentinH:
@all-contributors please add @FillPower1 for code

PR #125, opened 2020-05-07
Manage mobile gestures from props

TiTnOuK:
Adding the prop `gesture` to be able to activate or not zoom and rotation gestures on mobile.
The prop take an object as value : { zoom: boolean, rotation: boolean } and by default all gestures are enabled.


ValentinH:
Thanks for making the effort to submit a PR.

I'm wondering why do you need this? Indeed, if you don't set the `onZoomChange` and `onRotationChange`, you should have the same result. Am I missing something?

TiTnOuK:
Oh you're right, I misunderstood the use of these callbacks. I thought it was always called after a change of zoom or rotation via gesture or not.

I close this pull request because useless indeed. Thanks for the precision :)

ValentinH:
Indeed, on mobile, the only way to change zoom or rotation is via gestures :)

PR #127, opened 2020-05-12
Remove Emotion and add inline css

stefanoruth:
This i my proposal at the moment to remove emotion, i would very much like to hear what you think about it. :)

It inlines the needed css inside the bundle, and i have added an option to disable the inline css, so people can add it to their own css file, and bundle it the way they want it.

I havent added any test to the new helper function let me know if you want that too,

I would also like to take a look at the tslib, but havent done anything about it yet (not sure if it can make bundle smaller or not yet)

In case of merging you are more than welcome to squash all the commits


ValentinH:
Thanks for this PR.

However, using `import cssStyles from './styles.css'` is only possible in Webpack. I guess it works if you run the demo locally, however if you look at the codesandbox that was automatically created to point on your compiled version, it doesn't work.

stefanoruth:
Yeah i can see it uses a dependency from node_modules in the compiled code which it shouldn't, i will take another look at this ;)

stefanoruth:
But what do you think of letting the code inject at styletag with the styles next to the component it self?

ValentinH:
> But what do you think of letting the code inject at styletag with the styles next to the component it self?

I find it awkward to be honest, I have never seen this before.
I think that if we want to use plain CSS, we should ask the developer to import it explicitly. 

Please see https://dev.to/stereobooster/what-is-the-best-way-to-distribute-styles-with-npm-package-m5f

stefanoruth:
Yeah i agree it a bit awkward also why id added the option to disable it, the article you are linking to also suggest some other css-in-js libs which has alot smaller size fx, https://bundlephobia.com/result?p=linaria@1.3.3 Which we could implement instead of emotion, but that leads to some other complications with SSR, but yeah exposing a css file to import proberly is the best solution even tho it gives some more work to the user of the package

ValentinH:
I didn't know linaria. If it works I'd be happy to use it in this library!

stefanoruth:
Truth be told i have never heard of linaria before that article

bili.config.ts
```diff
@@ -8,10 +8,16 @@ const config: Config = {
     format: ['cjs', 'umd', 'umd-min', 'esm'],
     moduleName: 'ReactEasyCrop',
     sourceMap: true,
+    extractCSS: false,
```
ValentinH, line 11: could you explain a bit what these 2 new options are doing please ? 🙂

examples/src/index.tsx
```diff
@@ -60,6 +60,7 @@ class App extends React.Component<{}, State> {
   render() {
     return (
       <div className="App">
+        <style>{styles}</style>
```
ValentinH, line 63: Why do we have to change the example? I think we should aim to keep the previous version working.

If I understand correctly, this change is not needed.

ValentinH:
Just tried the sandbox and it looks good to me, really cool stuff. I need more time to test this carefully and really understand the implications :).

I mainly want to read the build output :)

bili.config.ts
stefanoruth: By default the built tool bili extract css imports to its own files and output then in the dist folder, this disabled that feature,

After disabling it, the bili rollup configuration is set to automaticly inject the style tag in the head of the dom, this postcss config disables that, we needed to disabled that feature else the rollup script would end up inside the bundle

examples/src/index.tsx
stefanoruth: We actually dont have to change this, main reason was that i needed to change the webpack config to inline the css, and so this would fail because it also got inline, but i can create 2 different webpack rules for the src and example so we can keep the current behavoir and dont need to change the example ;)

stefanoruth:
That totally understandable, let me know if you want me to change anything ;)

ValentinH:
@stefanoruth I've checked it locally and I like the approach. However, I would change one thing: I would inject the style tag in the `<head>` instead of in the middle of the body.

This can be done in `componentDidMount()` with:
```ts
const styleTag = document.createElement('style')
styleTag.setAttribute('type', 'text/css')
styleTag.innerHTML = cssStyles
document.head.appendChild(styleTag)
```

Also I don't think we need the `disableInlineCss`  as without it the library cannot be used. I would release it like this without any new options. If there are some people willing to be able to manually inject the style, this could be added later. 


What do you think?

stefanoruth:
Move it to head now with the code you posted, also removed the `disableInlineCss` 😄  but i still think it would be a good idear to include somekind of prop to this the autoinjection, and let people add it to their own css setup. Both in terms of people being able add the style to their caching setup and add it to their own css processors, to their can any the of the autoprefixes that they have configured in their setup

ValentinH:
The more I think about it the more I agree :)

I quickly tried tweaking the config yesterday to also export the css as a plain file but didn't find a way so far. One possible approach would be to copy the src one to the dist folder with a simple bash command.

However, do you think we should create another export file without the inline CSS? This might be a too advanced optimisation for now so we could only start with just providing the CSS file and document how to use it manually.

stefanoruth:
Hmm i have a few options right now that i am considering:

Option A: Only export a css file and document how to use it by importing it. (This is definitely a breaking change)

Option B: Keeping the inline plus exporting a css file with documentation, but adding a prop to disable the inline (Could not be a breaking change, but could be used as a step towards Option A or permanent) (Exporting could as you suggested just be bash copy/paste)

Option C: Only having the inline styles + a prop to disable it, and letting people just copy paste the styles from the src folder into their project.

I am most for Option A but that being a breaking change might leave option b as the better choice.

ValentinH:
Thanks for detailing the options :)

I'm more in favour of option B for 2 reasons:
- avoid breaking changes (even if I'll still bump a major version just in case)
- not having to import the CSS manually is handy.

stefanoruth:
Added option b now:

-Copied the css and rename the file to what ever format bili build system normally would have called a compiled css file.
- Added a props called `disableDefaultStyles` to turn of the inject of styles in head, not sure about the naming of the props

Missing: some documentation of the prop and how to import the css file

src/index.tsx
```diff
@@ -46,6 +46,7 @@ type Props = {
   restrictPosition: boolean
   initialCroppedAreaPixels?: Area
   mediaProps: React.ImgHTMLAttributes<HTMLElement> | React.VideoHTMLAttributes<HTMLElement>
+  disableDefaultStyles?: boolean
```
ValentinH, line 49: I would suggest: `disableAutomaticFileInjection`

stefanoruth:
Changed the name of the props + add some documentation to the readme file, it seems like my code editor change some more of the structure of the file, let me know it you want me to try and revert the formatting

ValentinH:
I'm sorry but I didn't mistake when suggesting the name 😬. I wanted to say `disableAutomaticStylesInjection` 😅

One thing I thought about yesterday evening is if injecting at the end of the head is not breaking the support of custom classnames passed via props. I need to check.

stefanoruth:
Ill update it :), thats fair enough ;)

ValentinH:
Thank you really much! I'm going to publish this as version 3.0.0 as soon as NPM is back: https://status.npmjs.org 😅

ValentinH:
@allcontributors please add @stefanoruth for code and ideas

ValentinH:
Release published: https://github.com/ricardo-ch/react-easy-crop/releases/tag/v3.0.0 🎉

Issue #132, opened 2020-05-15
Logarithmic Zoom Delta on Mouse wheel


jamespedid:
## Are you making a feature request?
- Please describe your use case from user journey point of view, e.g.:
The scrolling of the zoom parameter is not linear due to the nature of zooming math. Exponentials and logarithms should be used for proper effect.

When setting a min zoom of 1/10 and a max zoom of 10, the mouse wheel of the zoom is 
 unbalanced in terms of the amount of zoom that you get per 'delta y'; when the zoom is around 1, the zoom changes much faster than when around the maximum and minimum values.

- If you have ideas how to implement your new feature, please share!

The reason for this is due to the calculation of new zoom. Instead, if you use proper base ten logarithm in the coordinates, then the zoom would be linear.

`newZoom = zoom - deltaY * zoomSpeed / 200`

 should instead become

```
zoomTValue = Math.log10(zoom);
minZoomTValue = Math.log10(minZoom);
maxZoomTValue = Math.log10(maxZoom);
newZoomTValue = zoomTValue - deltaY * zoomSpeed * (maxZoomTValue - minZoomTValue) // maybe scale it by the 200 here
newZoom = Math.pow(10, newZoomTValue);
```

ValentinH:
Thanks for the feature request. I think it makes sense. Would you like to submit a PR for this?

jamespedid:
I'll give it a crack a little later tonight.

jamespedid:
Ended up doing it now while I had a bit of free time waiting for a meeting. Take a look. I'd personally recommend making it the default behavior because I think it gives a more natural scrolling effect.

ValentinH:
Closing this now as no one else requested this feature. We are not going to implement this.

PR #133, opened 2020-05-15
added logarithmic scrolling option for a smooth scrolling option.

jamespedid:
Addresses #132 

Added an option that defaults to false for using logarithmic scrolling.

When enabled, logarithms and exponentials are used to calculate a new zoom value that gives a linear scrolling effect.

For an obvious comparison when testing, use a MIN_ZOOM=1/10 and MAX_ZOOM=10 to notice the obvious differences.

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -47,6 +47,7 @@ type Props = {
   initialCroppedAreaPixels?: Area
   mediaProps: React.ImgHTMLAttributes<HTMLElement> | React.VideoHTMLAttributes<HTMLElement>
   disableAutomaticStylesInjection?: boolean
+  logarithmicScrolling: boolean
 }
 
 type State = {
@@ -72,6 +73,7 @@ class Cropper extends React.Component<Props, State> {
     zoomSpeed: 1,
     restrictPosition: true,
     zoomWithScroll: true,
+    logarithmicScrolling: false,
   }
 
   imageRef: HTMLImageElement | null = null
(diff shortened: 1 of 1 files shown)


src/index.tsx
```diff
@@ -329,10 +331,21 @@ class Cropper extends React.Component<Props, State> {
...
+    const maxZoomY = Math.log10(this.props.maxZoom)
+    const newZoomY = zoomY - (deltaY * this.props.zoomSpeed * (maxZoomY - minZoomY)) / 200
+    return Math.pow(10, newZoomY)
+  }
+
   onWheel = (e: WheelEvent) => {
     e.preventDefault()
     const point = Cropper.getMousePoint(e)
-    const newZoom = this.props.zoom - (e.deltaY * this.props.zoomSpeed) / 200
+    const newZoom = this.calculateNewZoom(e.deltaY)
```
jamespedid, line 348: might want to instead divide here by 200 so that the deltaY for any other way of calling calcluateNewZoom isn't scaled by this pixel value, and is instead scaled in the natural zooming unit.

ValentinH:
Do you still want to continue on this PR or can I close it?

jamespedid:
Sorry, I got pulled onto other things. I'm gonna close the PR but if you want to investigate this further (or feel it's no longer necessary), then feel free. My goal with this change was to provide a linear sliding scale when moving from sizes that would feel smooth, and logarithms seemed to be the answer. But I have no idea how to tune these in a way that would make the most sense for this project and it's goals.

Issue #123, opened 2020-05-02
Invariant violation: Maximum update depth exceeded when using crop size


jmorey28:
## Are you reporting a bug?
yes

- Please create a CodeSandbox that demonstrates your issue. 
https://codesandbox.io/s/zen-chaplygin-t2jbk

- Provide the steps to reproduce the issue, e.g.:

1. using version 2.1.1 add cropSize prop to component, with 200 for width and height (seems to be with any values)

  **Observed behaviour:** Maximum update depth exceeded error
  **Expected behaviour:** Should allow cropSize without error

ValentinH:
Please see #118

jmorey28:
great thanks!

PR #134, opened 2020-05-18
Compare cropsize properties individually instead of object references.

dvail:
I also ran into the issue described here: https://github.com/ricardo-ch/react-easy-crop/issues/123, which even though that issue was closed has an easy fix.

Automated and manual tests passed, and I was able to verify the infinite loop crash no longer occurs when setting the `cropSize` object directly as a component prop in examples/index.tsx.


ValentinH:
Thanks for this! I've merged it and will publish it probably tomorrow (away from computer right now).

ValentinH:
@allcontributors please add @dvail  for code

PR #146, opened 2020-06-11
Add option for flipping the image

ajlende:
I'm working on a feature in https://github.com/WordPress/gutenberg/pull/22959 that would benefit from having a flip option. This PR adds options for vertical and horizontal flipping. They just get applied in the CSS `transform` property like the rotations.


ValentinH:
Thanks for this PR, I like the feature.

I'm wondering 2 things: 
- should we expose a callback to compute the transformation string to cover any other uses cases? That would let us add only one prop that could cover any other use cases regarding custom transform operations.
- do you know how to extend the canvas croping logic that is used in the example to display the output? If we demo the flip feature, it needs to be supported in the output.

ajlende:
Good questions!

> should we expose a callback to compute the transformation string to cover any other uses cases? That would let us add only one prop that could cover any other use cases regarding custom transform operations.

The only other types of transformations left that can be done with CSS transform are shearing/skewing and perspective transformations, so there wouldn't be a large number of additions if you wanted to cover the remaining use cases. Maybe for some people it could be useful to provide the matrix transformation, but unless you're just passing the matrix through from another source, it's always more readable to use the named transformations instead.

> do you know how to extend the canvas cropping logic that is used in the example to display the output? If we demo the flip feature, it needs to be supported in the output.

I should be able to get that to work. I also realized that I still have to test this with the rotation gestures because I might have to inverse the direction of rotation when a flip is applied.

ajlende:
I wasn't sure how you wanted to handle the demo since it was pointing at a published version, so I updated it to pull local files from `dist` instead. You'll just have to build before installing with this setup.

ajlende:
And some more documentation might be needed for handling the rotation change that comes with flipping. Not sure if there's a way to automatically handle it because the state lives in the parent component.

https://github.com/ajlende/react-easy-crop/blob/33a9293cb9acb98990912bcc50bbe7850d0dbf32/examples/src/index.tsx#L85-L93

ajlende:
@ValentinH did you have any thoughts about having to set the rotation state on flip? Is there anything else that you'd like me to change?

ValentinH:
Hey, sorry I didn't see you already updated the demo and was waiting for it ^^

Indeed, the "extra setState on flip" is quite annoying. If I'm not mistaken, if we expose a callback to explicitly set the `transform` string, this would not be needed, right?

ajlende:
Unfortunately adding the transform callback won't fix the problem of having to maintain the rotation when flip changes because that state already lives in the parent. I threw together a demo of the transform prop on [another branch](https://github.com/ricardo-ch/react-easy-crop/compare/master...ajlende:transform-prop) as an example.

Something that I hadn't considered earlier was how, given a new `transform` prop, you'd push the responsibility for figuring out the proper order on the user. I suppose it's your decision if that added responsibility is a feature or a drawback for `react-easy-crop`. I made a couple GIFs to explain the order thing because it's hard to describe.

**Rotate/Flip then Translate** flips over the image's center. Notice the right edge stays the same.

rotate-then-translate

**Translate then Rotate/Flip** flips over the canvas's center. Notice the puppy's eye stays in the same place in the center.

translate-then-rotate

Typically, the order of operations would be scale, rotate, translate (i.e. `transform: translate() rotate() scale()` because operations are applied right to left) but you can change the effect if you understand how it all works.

ValentinH:
TIL: you can change the order of the transformations and it changes the behavior! 😲 Really cool

I've created a PR from your branch (#149) so I can test it in Codesandbox. At first, I was thinking about using a function but they way you did it with the `transform` string is pretty cool. I need to test it but I think I prefer this approach.

Regarding the order, I don't think this is an issue as the actual cropping implementation is also left to be done by the developer. It just needs to be consistent.

ValentinH:
I just tried your branch and here's a demo: https://codesandbox.io/s/crazy-blackburn-eur15

You can actually do the `360 - rotation` trick directly when computing the transform string
image


However, I'm not sure to understand why you want to use this logic, i.e. `360 - rotation`. 🤔

ajlende:
The `360 - rotation` trick is used to keep the slider in range when you flip so you can rotate back after you've flipped. Otherwise trying to rotate back might mean you'd have to go out of range of the slider. Because it's for the slider, it had to be in the `setState`.

ajlende:
The order can seem a little weird at first. You have to think about the origin that the transformations are happening about. In the CSS case the origin is in the center of the div. Every transformation happens about the same origin. I found a good diagram that shows it pretty well.

order of operations

ValentinH:
>  you'd have to go out of range of the slider

I don't see this behaviour in the sandbox I linked above 🤔 It's not a big issue as this will not be implemented as part of the library but I'm still curious to understand the issue 😄

ValentinH:
So if you apply the things that you did on the other branch, I'm fine merging it ;)

ValentinH:
This can be closed as it was implemented differently in #150

PR #149, opened 2020-06-23
Transform prop

ValentinH:
I'm creating a PR just to be able to quickly test this in Codesandbox :)


README.md
```diff
@@ -108,6 +108,7 @@ This component requires some styles to be available in the document. By default,
...
 | [`onCropComplete`](#onCropCompleteProp) | Function                                                                            |          | Called when the user stops moving the media or stops zooming. It will be passed the corresponding cropped area on the media in percentages and pixels                                                                                                                                                       |
+| `transform`                             | string                                                                              |          | CSS transform to apply to the image in the editor. Defaults to `translate(${crop.x}px, ${crop.y}px rotate(${rotation}deg) scale(${zoom})` with variables being pulled from props.                                                                                                                           |
```
ValentinH, line 111:
```suggestion
| `transform`                             | string                                                                              |          | CSS transform to apply to the image in the editor. Defaults to `translate(${crop.x}px, ${crop.y}px) rotate(${rotation}deg) scale(${zoom})` with variables being pulled from props.                                                                                                                           |
```

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "react-easy-crop",
-  "version": "3.0.1",
+  "version": "3.1.0",
```
ValentinH, line 3: You should not update the version number in the PR. This will done on `master` by using `np`.

ajlende: I suppose it isn't a problem for this PR because the version between the local package and the dependency in docs was already different, but when working with different versions I could only get yarn to pick up the change when the version number changed. Should I also revert the lockfile change (32566a5539a12edf2843575b4b6873c131982679) and then open another PR after publishing to update the docs with the new version then?

ValentinH: Yes, updating the version in the doc is quite annoying but is done a manual commit after the new version is released.

The real version of the package should not be increased in PR, otherwise, it'll prevent me from publishing the new version with the default pipeline.

ajlende: 👍 done!

ValentinH:
This looks good to me. Would you like to open a new PR so that you are listed as the PR creator or I can merge it like this?

PR #150, opened 2020-06-24
Transform prop and flip demo

ajlende:
Copied over from #149 so I'm listed as the PR creator. I had to make a new branch, but the commits are the same.


ValentinH:
Published with version 3.1.0 🎉

Issue #152, opened 2020-07-09
for a portrait image on rotation moves crop Rect outside image


SamirNvi:
Hi Valentin/ support ,

Thanks for this wonderful free library. It works nice for me.
I am facing a issue for portrait images. I Have portrait images with aspect ratio set to say 0.66. when i rotate the image to 90 degree my crop rect doesnt change and it stays outside image . i am setting the initial crop area as center cropped area of image.

pls help me to resolve this issue.

![image](https://user-images.githubusercontent.com/25320256/87024179-66db8700-c1f6-11ea-9f73-9277118578ea.png)

after rotate
![image](https://user-images.githubusercontent.com/25320256/87024409-b457f400-c1f6-11ea-9f6f-534921195aa3.png)

ValentinH:
Thanks for the kind words.

It looks like a bug indeed, could you please create a code sandbox showing the issue?

ValentinH:
After checking this bug, it's related to #151. Basically, vertical aspects/images are not properly handled and the internal `getCropSize()` needs to be improved.

ValentinH:
This should be fixed by #157, could you please test it (using the sandbox linked in the PR) and confirm?

ValentinH:
Fix release with v3.1.1

PR #157, opened 2020-07-18
fix(crop size): handle vertical ratio properly

### LINKED ISSUE #151: Unable to zoom out on rotated image and aspect ratio
Rotating an image and adjusting the aspect ratio to match the rotation makes the crop area smaller than the image and unable to be zoomed out.

image

CodeSandbox for the issue: https://codesandbox.io/s/react-easy-crop-image-aspect-fyie3

Steps to reproduce the issue:

  1.  Rotate image 90 degrees
  2.  Set aspect ratio to match the rotated image

**Observed behaviour:** The crop area is smaller than the image even though zoom is set to 0.

**Expected behaviour:** The crop area is the same size as the image, as in the initial state.

ValentinH:
Fixes #151 #152

We are now using the container size to compute the crop size. Thanks to this, we can better use to available size for the crop area.

@ajlende Could you please confirm that this fix is working for your use case? Here's your example running the library version from this PR: https://codesandbox.io/s/react-easy-crop-image-aspect-lwpee
You can use this version of the package to test locally: "https://pkg.csb.dev/ricardo-ch/react-easy-crop/commit/4f8563a2/react-easy-crop"

### DIFF
--- a/src/helpers.test.ts
+++ b/src/helpers.test.ts
@@ -3,24 +3,34 @@ import * as helpers from './helpers'
 describe('Helpers', () => {
   describe('getCropSize', () => {
     test('when media width is higher than the height based on the aspect', () => {
-      const cropSize = helpers.getCropSize(1200, 600, 4 / 3)
+      const cropSize = helpers.getCropSize(1200, 600, 1000, 600, 4 / 3)
       expect(cropSize).toEqual({ height: 600, width: 800 })
     })
     test('when media width is smaller than the height based on the aspect', () => {
-      const cropSize = helpers.getCropSize(600, 1200, 4 / 3)
+      const cropSize = helpers.getCropSize(600, 1200, 1000, 600, 4 / 3)
       expect(cropSize).toEqual({ height: 450, width: 600 })
     })
-    test('when media dimensions exactly match the aspect', () => {
-      const cropSize = helpers.getCropSize(800, 600, 4 / 3)
+    test('when media dimensions exactly match the horizontal aspect', () => {
+      const cropSize = helpers.getCropSize(800, 600, 1000, 600, 4 / 3)
       expect(cropSize).toEqual({ height: 600, width: 800 })
     })
+    test('when media dimensions exactly match the vertical aspect', () => {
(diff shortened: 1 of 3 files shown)


ajlende:
Seems like it's closer than before, but when the window is really wide I still see the issue, even in the sandbox. It may be because when the image is rotated, the CSS transform doesn't recalculate the size of the element, so measurements are still based off the non-rotated width and height?

image

ValentinH:
I'm not sure to understand which issue you see. What output would you expect?

The crop area size looks correct to me. If you are talking about the image being taller than the screen, this is indeed because we are using a transformation to rotate it.

ajlende:
Notice that the zoom slider. It is zoomed out as much as possible. You should be able to zoom out further so that the whole image fits within the crop area.

ValentinH:
I see what you mean. Basically, you need to allow a zoom value lower than 1 in this case (probably equals to the aspect ratio?).

The image positioning is not handled by JavaScript, we only let the browser fit the image in the container. As mentioned above, when we rotate the image using a transformation, then this automatic positioning let the image flow outside of its container.

I don't see (at the moment) how to address this as part of the library without changing too many things.

ajlende:
> Basically, you need to allow a zoom value lower than 1 in this case.

I'd prefer that a zoom value of 1, an aspect ratio that matches the image (even when rotated), and `restrictPosition` set to true to always cover the entire image. That is, a value of less than one would only serve as to add padding around the image, if allowed.

> I don't see (at the moment) how to address this as part of the library without changing too many things.

That's okay, this is a step in the right direction. And it seems to fix #152 for me. 👍 

I'll start playing around with a solution for #151, but it sounds like it might end up needing to change a lot, so we'll see how it goes.

ValentinH:
> I'd prefer that a zoom value of 1, an aspect ratio that matches the image (even when rotated), and restrictPosition set to true to always cover the entire image. That is, a value of less than one would only serve as to add padding around the image, if allowed.

Yes, it makes sense. However, with the way the image is rotated today (transformation), this is tricky. If you come up with a PR for this, I'll be happy to review.

ValentinH:
Released under v3.1.1

PR #179, opened 2020-09-23
fix(crop-size): Add function callback prop to access crop size changes

### LINKED ISSUE #178: Emit cropSize changes
**Is your feature request related to a problem? Please describe.**
Can not access cropSize to know the width/height of the cropping area when window/container is resized or when different sized images are used.

**Describe the solution you'd like**
Would like to know the cropping area's width and height to be able to add some overlayed stylings related to the cropping area, e.g. animations. A function callback prop that would be called when the cropSize width or height changes. Something like `onCropSizeChanged` can be passed and it would be called with the updated `cropSize` whenever it is changed.

**Describe alternatives you've considered**
A use case, we are not setting the cropSize manually since we want to use the default cropping area resizing rather than calculating it ourselves. Because of this the cropping area will change depending on the container or the image dimensions. We tried using the imageDimensions from onMediaLoad to calculate the cropSize but if the window is resized it could give an incorrect value when the cropper is resized.

calemon:
Fixes #178

Add a callback prop to use when a change between previous cropSize width or height is calculated in `getCropSize`.


src/index.tsx
```diff
@@ -228,6 +229,31 @@ class Cropper extends React.Component<Props, State> {
     }
   }
 
+  getEmitCropSize = (
```
ValentinH, line 232: I'm not a big fan of this `getEmit` pattern. Indeed, when calling this method, we are not sure that the returned cropSize will be actually used in the component. In the future, we could add some logic before calling `setState()` on line 228 and we could forget about this.

I'd rather to the checks right before setting the state. Even, if the `cropSize` might come from the props, I don't think this is an issue as the `onCropSizeChange` callback is still doing what is is supposed to do 🙂

ValentinH (commented):
Thanks for creating this PR, I'm OK with the feature proposal. Just one remark regarding the approach.

src/index.tsx
calemon: Very true, great call... I've updated the PR, moving the checks for cropSize changes to right before setting the state of the calculated cropSize value. If there is a difference between width or height, the callback prop will be called.

ValentinH (approved):

ValentinH:
I'll merge and release a new version tomorrow morning. I'm not on my computer right now ...

ValentinH:
Published on 3.2.0

PR #181, opened 2020-10-12
Export props to allow other libraries to reference it

labithiotis:
Pretty simple PR to export props to allow them to be reference by other libraries.

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -12,7 +12,7 @@ import {
 } from './helpers'
 import cssStyles from './styles.css'
 
-export type EasyCropperProps = {
+export type CropperProps = {
   image?: string
   video?: string
   transform?: string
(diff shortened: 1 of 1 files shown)


src/index.tsx
```diff
@@ -12,7 +12,7 @@ import {
 } from './helpers'
 import cssStyles from './styles.css'
 
-type Props = {
+export type EasyCropperProps = {
```
ValentinH, line 15: As the component is named Cropper, I would name this CropperProps.

labithiotis: @ValentinH Can do, but thought due to NPM name it would be best to relate to that i.e `react-easy-crop` therefor 'EasyCropper`

ValentinH: Cropper is the name used in the docs so I'd use this :)

ValentinH:
Thanks for this PR. It has been published under v3.2.1 🎉

ValentinH:
@all-contributors please add @labithiotis for code

PR #188, opened 2020-10-28
Recompute crop position when crop changes from outside props

nihey:
My problem: I am changing the `crop` prop from outside to achieve horizontal or vertical centering through buttons, but changing it does not trigger `onCropComplete` callback, this way the preview is shown as if the crop is centered, but the final crop (that depends on the `croppedAreaPixels` from `onCropComplete`) is not correct.

This video demonstrates the issue: https://streamable.com/i4v1xr

The source code for this Crop Dialog can be seen here: https://github.com/KasselLabs/ejector/blob/main/src/components/CropDialog.js

An example of this behavior can be tested here: https://ejector.kassellabs.io/

The change in this PR fixes this issue for me, is everything correct with it?

Keep up this great project!

### DIFF
--- a/examples/src/index.tsx
+++ b/examples/src/index.tsx
@@ -19,6 +19,7 @@ type State = {
   showGrid: boolean
   zoomSpeed: number
   restrictPosition: boolean
+  croppedArea: Area | null
 }
 
 class App extends React.Component<{}, State> {
@@ -33,6 +34,7 @@ class App extends React.Component<{}, State> {
     showGrid: true,
     zoomSpeed: 1,
     restrictPosition: true,
+    croppedArea: null,
   }
 
   onCropChange = (crop: Point) => {
@@ -41,6 +43,7 @@ class App extends React.Component<{}, State> {
 
   onCropComplete = (croppedArea: Area, croppedAreaPixels: Area) => {
     console.log(croppedArea, croppedAreaPixels)
+    this.setState({ croppedArea })
   }
 
   onZoomChange = (zoom: number) => {
@@ -111,6 +114,35 @@ class App extends React.Component<{}, State> {
               Flip Vertical
             </label>
           </div>
+          <button
+            id="horizontal-center-button"
+            onClick={() => {
+              this.setState({
+                crop: { ...this.state.crop, x: 0 },
+              })
+            }}
+          >
+            Center Horizontally
+          </button>
+          <div>
+            <p>Crop Area:</p>
+            <div>
+              {['x', 'y', 'width', 'height'].map((attribute: string) => {
+                if (!this.state.croppedArea) {
+                  return null
+                }
+
+                return (
+                  <div key={attribute}>
+                    {attribute}:
+                    <b id={`crop-area-${attribute}`}>
+                      {Math.round(this.state.croppedArea[attribute])}
+                    </b>
+                  </div>
+                )
+              })}
+            </div>
+          </div>
         </div>
         <div className="crop-container">
           <Cropper
@@ -126,6 +158,7 @@ class App extends React.Component<{}, State> {
             onCropChange={this.onCropChange}
             onRotationChange={this.onRotationChange}
             onCropComplete={this.onCropComplete}
+            onCropAreaChange={this.onCropComplete}
             onZoomChange={this.onZoomChange}
             onInteractionStart={this.onInteractionStart}
             onInteractionEnd={this.onInteractionEnd}
--- a/examples/src/index.tsx
+++ b/examples/src/index.tsx
@@ -127,7 +127,7 @@ class App extends React.Component<{}, State> {
           <div>
             <p>Crop Area:</p>
             <div>
-              {['x', 'y', 'width', 'height'].map((attribute: string) => {
+              {(['x', 'y', 'width', 'height'] as const).map((attribute) => {
                 if (!this.state.croppedArea) {
                   return null
                 }
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -9,6 +9,7 @@ import {
   getCenter,
   getInitialCropFromCroppedAreaPixels,
   classNames,
+  debounce,
 } from './helpers'
 import cssStyles from './styles.css'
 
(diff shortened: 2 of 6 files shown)


ValentinH:
Hello @nihey and thanks for this PR. I understand the use case but this change is kind of a breaking change: now the `onCropComplete` is called on every mouse move event. I tested the performances and this is fine in the simple example. However some application might have heavy computations run on `onCropComplete`.

One easy solution I see is to add an option to enable this behaviour. However, the `onCropComplete` would still be not correctly named anymore...

Do you have any idea?

nihey:
Hi @ValentinH 

I understand the problem, this could indeed cause a problem on people that auto crop the images as soon as the `onCropComplete` returns.

> One easy solution I see is to add an option to enable this behaviour. However, the onCropComplete would still be not correctly named anymore...

I could add another callback called `onCropAreaChanged` and use it instead of the `onCropComplete` to avoid the naming not being correct.

> Do you have any ideas?

I thought about another alternative:

Adding a `debounce` to this specific `this.emitCropData` that I've added could reduce the amount of CPU stress for this case.

Let me know what you think

ValentinH:
I like both the  `onCropAreaChanged` and  `debounce` logic. I think both could be combined.

We still to keep the existing behaviour of `onCropAreaChanged` though, so I think you could add an extra parameter to `emitCropData(isCropComplete = false)`, what do you think?

nihey:
> I like both the onCropAreaChanged and debounce logic. I think both could be combined.

That looks good, I'll make both of these changes and will be back with the fix.

> We still to keep the existing behaviour of onCropAreaChanged though, so I think you could add an extra parameter to emitCropData(isCropComplete = false), what do you think?

This seems like the way to go for me too. `this.emitCropData(isCropComplete = true)` could be the default to keep everything as it is today, passing `this.emitCropData(false)` would trigger `onCropAreaChanged` without triggering `onCropComplete`.

I can also update the documentation and possibly add some tests for this new prop. I haven't used Cypress yet, but it is something I'm willing to use on future projects, it seems like a great e2e testing lib.

ValentinH:
Yes Cypress is great! We are doing very little usage of it here though as our test case are simple and not 100% reflecting what a user can do and see. Indeed, we need to directly observe CSS properties which is not ideal...

src/index.tsx
```diff
@@ -424,10 +434,43 @@ class Cropper extends React.Component<CropperProps, State> {
       this.props.rotation,
       this.props.restrictPosition
     )
-    this.props.onCropComplete &&
+  }
+
+  emitCropAreaChange = () => {
```
nihey, line 439: @ValentinH I've started coding this solution following our `emitCropData(isCropComplete = true)` discussed function, but in the end, it felt more natural to use this specific `emitCropAreaChange` function, that calls the `debouncedEmitOnCropComplete` function.

This way I do not need to manually call:

```javascript
this.emitCropData(false);
this.debouncedEmitCropData();
```

Inside `componentDidUpdate`. This would also trigger `onCropAreaChange` twice every time if the crop is controlled from an outside source.

Let me know what you think.

examples/src/index.tsx
```diff
@@ -111,6 +114,35 @@ class App extends React.Component<{}, State> {
...
+                  return null
+                }
+
+                return (
+                  <div key={attribute}>
+                    {attribute}:
+                    <b id={`crop-area-${attribute}`}>
+                      {Math.round(this.state.croppedArea[attribute])}
+                    </b>
+                  </div>
```
nihey, line 141: @ValentinH I've also added this little HTML by the top of the example:

Screenshot from 2020-10-31 23-52-46

This made Cypress testing simpler, as I can actually debug and access the real output from `onCropComplete` and `onCropAreaChange` functions. Does it make sense?

src/index.tsx
ValentinH: I like having 2 distinct `emitCropData()` and  `emitCropAreaChange()`.

However, I see a few small issues here:
- `debouncedEmitOnCropComplete` should not be called inside `emitCropAreaChange` as this is changing the existing behaviour. This new function should only call the new `onCropAreaChange`.
- I think we should debounce `onCropAreaChange` and not `onCropComplete`
- even by doing the 2 updates above, there is a small issue with the debounced callback when zooming with the mouse scroll. Indeed, the debounced call is called 300ms after the last non-debounced call, but with outdated values. I don't know if what I mean is clear but you can see the examples output you have added, the data changed 300ms after the last scroll and is not correct.

If this is fine for you, I would simply remove the debounce logic as there is not performance issues at the moment, let's not optimise imaginary issues. 🙂

src/types.ts
```diff
@@ -16,7 +16,7 @@ export type Point = {
 }
 
 export type Area = {
-  width: number
+  [width: string]: number
```
ValentinH, line 19: why are you changing this? 🤔

src/index.tsx
nihey: > If this is fine for you, I would simply remove the debounce logic as there is not performance issues at the moment, let's not optimise imaginary issues. :slightly_smiling_face:

I think it would be the best thing to do too :slightly_smiling_face: . This way I can only focus on solving the specific problem of not having a `onCropAreaChange` callback to receive when adjusting the crop through external buttons. I will avoid a lot of unnecessary complexity and dealing with those issues pointed above.

src/types.ts
nihey: I'm sorry if it is a mistake, I'm not too familiar with Typescript.

I was getting a warning due to this [`this.state.croppedArea[attribute]`](https://github.com/ricardo-ch/react-easy-crop/pull/188/files#diff-64c7378dfabef2bf61425c20979c3762672c25a7ccd9a90ddef13222476f25a8R139) property access. It seemed like Typescript wanted me to explicitly declare that the keys within the `Area` object were `string` typed. Would this be the best way to go with it?

ValentinH: I actually don't get this TS error when reverting this change. However I see what TS could complain, you can rewrite the `map()` line like this:
```ts
(['x', 'y', 'width', 'height'] as const).map((attribute) => {
...
})
```

nihey: @ValentinH Thanks for help, this worked perfectly, I've reverted this change back to the original state.

ValentinH (approved):
All good for me 🚀

ValentinH:
Published with v3.3.0: https://github.com/ricardo-ch/react-easy-crop/releases/tag/v3.3.0

PR #190, opened 2020-11-13
Fix document not defined error on server side rendering

digz6666:
Fix document not defined error on server side rendering.

Will fix following issue too:
https://github.com/nanxiaobei/antd-img-crop/issues/98

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -106,6 +106,8 @@ class Cropper extends React.Component<CropperProps, State> {
     }
 
     if (!this.props.disableAutomaticStylesInjection) {
+      if (typeof document === 'undefined') return;
+
       this.styleRef = document.createElement('style')
       this.styleRef.setAttribute('type', 'text/css')
       this.styleRef.innerHTML = cssStyles


src/index.tsx
```diff
@@ -106,6 +106,8 @@ class Cropper extends React.Component<CropperProps, State> {
     }
 
     if (!this.props.disableAutomaticStylesInjection) {
+      if (typeof document === 'undefined') return;
```
ValentinH, line 109: This code is inside `componentDidMount()` so it's never run on the server. 

The issue you linked doesn't show an error on this package but in the  `antd-img-crop` package.

ValentinH: Our demo website is actually using SSR so there is no issue on our side.

digz6666: I'm using ReactDOM.hydrate and the components are loaded and rendered on the server side and returns to client.
Adding this line is fixed the issue for me.

ValentinH: I don't think this is correct: we are already accessing the window object at line 100 so this would break as well on the server.

As I said,  componentDidMount doesn't run the server .

ValentinH: Are you running ReactDOM.hydrate on the server?

digz6666: Yes, I think ReactDOM.hydrate is called on the server.

ValentinH: Then, this is your issue. `react-dom`'s `hydrate is for the DOM so not for the server.
You should use  `react-dom/server` and the `renderToString()` function on the server.

digz6666: I see, I will check my code.

digz6666: I'm using server side rendering to solve the SEO issues.

digz6666: My server.js uses react-dom/server and renderToString and on the client side it uses react-dom and hydrate.

ValentinH: Yes so this is correct. So something else should be wrong in your config.
See this doc that explains that the only lifecycle method called during SSR is `componentWillMount()`: https://github.com/axios/axios/issues/1175

digz6666: Ok thanks, I'll close this pull request.

ValentinH: Good luck for the resolution 🙂

PR #198, opened 2020-12-20
Fixed wrong initial position when image was rotated

b-des:
Fixed issue when seting ```initialCroppedAreaPixels``` led to wrong position if image was rotated before;
Added support of rotated parent container, using ```transform``` css property.

### DIFF
--- a/src/helpers.ts
+++ b/src/helpers.ts
@@ -181,7 +181,8 @@ function getZoomFromCroppedAreaPixels(
 export function getInitialCropFromCroppedAreaPixels(
   croppedAreaPixels: Area,
   mediaSize: MediaSize,
-  cropSize?: Size
+  rotation: number,
+  cropSize?: Size,
 ): { crop: Point; zoom: number } {
   const mediaZoom = mediaSize.width / mediaSize.naturalWidth
 
@@ -189,11 +190,49 @@ export function getInitialCropFromCroppedAreaPixels(
 
   const cropZoom = mediaZoom * zoom
 
-  const crop = {
+  let crop = {
     x: ((mediaSize.naturalWidth - croppedAreaPixels.width) / 2 - croppedAreaPixels.x) * cropZoom,
     y: ((mediaSize.naturalHeight - croppedAreaPixels.height) / 2 - croppedAreaPixels.y) * cropZoom,
   }
+  crop = transformCoordinateByAngle(crop, -rotation * Math.PI / 180);
   return { crop, zoom }
+}
+
+/**
+ * Apply rotation formula to point and return new coordinate
+ * @param point
+ * @param radians
+ */
+export function  transformCoordinateByAngle(point: Point, radians: number){
+    return {
+        x: point.y*Math.sin(radians) + point.x*Math.cos(radians),
+        y: point.y*Math.cos(radians) - point.x*Math.sin(radians)
+    }
+}
+
+/**
+ * Return rotation property of transform style of given element
+ * @param target
+ */
+export function getRotationAngle(target: HTMLElement){
+    const obj = window.getComputedStyle(target, null);
+    const matrix = obj.getPropertyValue('-webkit-transform') ||
+        obj.getPropertyValue('-moz-transform') ||
+        obj.getPropertyValue('-ms-transform') ||
+        obj.getPropertyValue('-o-transform') ||
+        obj.getPropertyValue('transform');
+
+    let angle = 0;
+
+    if (matrix !== 'none')
+    {
+        const values = matrix.split('(')[1].split(')')[0].split(',');
+        const a = parseFloat(values[0]);
+        const b = parseFloat(values[1]);
+        angle = Math.round(Math.atan2(b, a) * (180/Math.PI));
+    }
+
+    return (angle < 0) ? angle += 360 : angle;
 }
 
 /**
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -10,6 +10,8 @@ import {
   getCenter,
   getInitialCropFromCroppedAreaPixels,
   classNames,
+  transformCoordinateByAngle,
+  getRotationAngle
 } from './helpers'
 import cssStyles from './styles.css'
 
@@ -20,6 +22,7 @@ export type CropperProps = {
   crop: Point
   zoom: number
   rotation: number
+  parentRotationAngle: number
   aspect: number
   minZoom: number
   maxZoom: number
@@ -49,6 +52,7 @@ export type CropperProps = {
   }
   restrictPosition: boolean
   initialCroppedAreaPixels?: Area
+  initialRotation?: number
   mediaProps: React.ImgHTMLAttributes<HTMLElement> | React.VideoHTMLAttributes<HTMLElement>
   disableAutomaticStylesInjection?: boolean
 }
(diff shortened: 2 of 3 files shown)


b-des:
@ValentinH,
My PR couldn't pass workflow checks. Please, advise what went wrong.

src/index.tsx
```diff
@@ -197,6 +202,7 @@ class Cropper extends React.Component<CropperProps, State> {
     const { crop, zoom } = getInitialCropFromCroppedAreaPixels(
       initialCroppedAreaPixels,
       this.mediaSize,
+      rotation - initialRotation,
```
ValentinH, line 205: Why do you need to substract `initialRotation` from `rotation`?

src/index.tsx
```diff
@@ -284,19 +290,20 @@ class Cropper extends React.Component<CropperProps, State> {
   }
 
   onDragStart = ({ x, y }: Point) => {
-    this.dragStartPosition = { x, y }
+    this.transformAngle = this.props.parentRotationAngle || getRotationAngle(this.containerRef!.parentElement!) * Math.PI / 180
```
ValentinH, line 293: I don't think that computing the parent element rotation should be built-in this application. The `parentRotationAngle` is enough for this advanced usage. Thus, we can remove the `getRotationAngle` function from the codebase.

ValentinH: Moreover, to be aligned with the existing `rotation` prop, I think this property should be using degrees too. Thus, `transformCoordinateByAngle` second argument would be degrees.

src/helpers.ts
```diff
@@ -181,21 +181,60 @@ function getZoomFromCroppedAreaPixels(
...
+  crop = transformCoordinateByAngle(crop, -rotation * Math.PI / 180);
   return { crop, zoom }
 }
 
+/**
+ * Apply rotation formula to point and return new coordinate
+ * @param point
+ * @param radians
+ */
+export function  transformCoordinateByAngle(point: Point, radians: number){
```
ValentinH, line 206:
```suggestion
export function  rotateCoordinates(point: Point, radians: number){
```

ValentinH (commented):
You committed the `package-lock.json` file that we don't use in this project, it needs to be removed.

The failing build comes from failing linting. You can run the checks locally with `npm run test`.

Regarding the content of the PR, I started to comment on the part dealing with the parent rotation. However, after trying it and thinking about it, I'd prefer to only add support for the `initialRotation`. 
Handling the parent rotation is more tricky and I don't think that it should be supported inside the library.

src/index.tsx
b-des: Regarding substraction ```initialRotation``` from ```rotation``` it's part of the logic of rotation and it might seem slightly tricky but it necessary to handle the angle at which the position was changed before save it to local storage or elsewhere. 

Deeper explanation: imagine we rotated image 90 degrees, changed position and save the area and the angle to store, after page reload we use the saved values to initialize cropper - it's need to set ```initialRotation``` to 90 and ```rotation``` to 0, it will work fine, ok... But then we decide to rotate image(using ```rotation``` prop) to 90 again - this is where the problem lies. ```initialRotation``` shouldn't apply in case when ```rotation``` equal to original angle(when the position was changed before saving) cause we already have the correct area for this angle.

Simply put, it's required to depreciate the effect when second time the same angle will be applied. 
I know it's tangled and I wrote not the best explanation, sorry for that :)

About parent rotation - it's really up to you, I just suggested the improvement cause I personally encountered with this problem that's why I thought it might be useful for other users.

src/index.tsx
b-des: Ok, I can just leave ```parentRotationAngle``` and eliminate ```getRotationAngle``` from the codebase. 
I do agree we need to keep the dimensions in consistence, ```parentRotationAngle``` must be translated to radian.

src/index.tsx
ValentinH: I just noticed something: can't we only use the `rotation` prop here?
If it's correctly set on load it would avoid the issue you mentioned I think.

b-des: I got you opinion, but I don't have idea how to do it without extra prop. Since the rotation formula was apllied to all points, we need to know what the angle it was when the position was exported - it's kind of "anchor" for further transformation. This is the reason why we need new one prop ```initialRotation``` 

Otherwise it's need to think on another solution if we want to avoid additional prop

ValentinH: My point is about initialising `rotation` with the value coming from local storage. I think it should do the job. However, the logic in `getInitialCropFromCroppedAreaPixels` should be adjusted to stop doing `- rotation`

b-des: Another solution could be wrap the image with div container and apply rotation to it. But it can lead to other bugs and unexpected dragging behavior

b-des: > 
> 
> My point is about initialising `rotation` with the value coming from local storage. I think it should do the job. However, the logic in `getInitialCropFromCroppedAreaPixels` should be adjusted to stop doing `- rotation`

I've tried to do it in such way but id didn't work. I'll try to come up with it again, but I'm afraid it could be so complex

ValentinH: I just tried it out and now I'm actually wondering if we even need a change: I tried storing the rotation in the local storage with the latest version and it looks to be working fine to me: https://codesandbox.io/s/react-easy-crop-demo-with-local-storage-73u6i?file=/src/index.js

Could you try it please and tell me if I'm missing something?

b-des: Hmm, you know, it works well on your demo. Looks like we caught some misunderstanding. Cause the person who opened that issue probably didn't apply the rotation when set ```initialCroppedAreaPixels```. From my side I considered the issue like "if image was rotated for ex. 90 deg, then shifted, and then during the next initialization with previous crop area and any other custom angle, the image should take the same position related to container but with different angle"

b-des: So let's deal only with ```parentRotationAngle``` I believe it could be helpful in some cases. For example to develop some constructor or wysiwyg editor, etc. What you think about it?

ValentinH: Thanks for checking the example, I'll share it on the original issue then.

Regarding `parentRotationAngle`, I am not a big fan of it to be honest, it looks really specific. Moreover, it doesn't cover all use cases: for example, when zooming with the mouse, the position is not correct anymore.

In order to be more generic, I would rather have a callback option like `getCoordinates()` to provide inversion-of-control. Thanks to it, the user would be able to do any kind of logic before the coordinates are used inside the library. What do you think?

b-des: Got it, no problem then :)
Obviously it might be useful for some cases, in general this is cool idea.
But now I'm not sure if that will help to fix dragging distortion when parent container was rotated or maybe even translated.

In order to not deliver not conceptual solutions in codebase, I can use raw code of the library in my project and modify it as I want to achieve needed effect, right?

ValentinH: Yes forking is always an option :)

b-des: So I think we can close this PR?

ValentinH: Yes I think this is better.

Thanks for the work and the nice communication. I hope you can contribute again in the future and have it merged :)

b-des: Thank you too. Maybe we would work in the future again. Have a nice coding )

foucdeg:
@b-des @ValentinH I'm having this same issue: setting `initialCroppedAreaPixels` does not yield the original `croppedAreaPixels` when there was a rotation as well.
Are there any solutions currently?

ValentinH:
Hi @foucdeg , I think you should open a new issue if your think it's needed. This PR was closed because the use case was already covered and the issue came from a misusage. You can check the comments on the files of the PR for more context.

PR #226, opened 2021-03-27
Updated README.md

sshahzaiib:
Added a new codesandbox link to the example of react-easy-crop with the image selected by the user.

After implementing the `react-easy-crop` in a project, I came across a case where we needed to crop a png with _transparent background_ but after the cropping, the transparent background was turning black and the logo title was also written in black.

In canvasUtils, the image creation mimetype is specified as `image/jpeg`, but converting the image to `image/png` clears the issue and converts the background to white by default.

`  return new Promise((resolve) => {
    canvas.toBlob((file) => {
      resolve(URL.createObjectURL(file))
    }, 'image/png')
  })`


ValentinH:
Thanks for sharing this solution. However, I prefer updating directly the existing sandbox instead of using a sandbox I cannot control (security reason).

PR #260, opened 2021-06-03
Add key prop to video source

MatixYo:
On source elements inside the videos, the key prop is missing causing errors.


src/index.tsx
```diff
@@ -548,7 +548,7 @@ class Cropper extends React.Component<CropperProps, State> {
               controls={false}
             >
               {(Array.isArray(video) ? video : [{ src: video }]).map((item) => (
-                <source {...item} />
+                <source key={item.type} {...item} />
```
ValentinH, line 551: Type is missing for the default case where video is not an array.

MatixYo: Switched to item.src as a key which should be unique.

ValentinH (approved):

ValentinH:
Thanks published as 3.5.1: https://github.com/ricardo-ch/react-easy-crop/releases/tag/v3.5.1

ValentinH:
@all-contributors please add @MatixYo  for code

PR #274, opened 2021-06-29
Refresh video source on component update

live-memories:
When video prop is dynamically updated <video><source> is not reflected since we have to call "load" API.


ValentinH:
Thanks for the contribution! 🎉

ValentinH:
This was released with v3.5.2

ValentinH:
@all-contributors please add @liveboom for code

PR #312, opened 2021-11-20
Change rotated crop area output

### LINKED ISSUE #240: After rotating a rectangular image, croppedAreaPixel range can be negative
**Describe the bug**
When rotating a rectangular image, the x and y coordinates of the croppedArea no longer align within the width and height values.

**To Reproduce**
1. Go to https://codesandbox.io/s/cocky-pare-zdopi?file=/src/index.js
2. Click on Rotate
3. Drag the image so that the top left corner of the image is aligned with the top left corner of the crop area
4. Notice that the croppedAreaPixels is now `{width: 473, height: 355, x: 163, y: -163}` when I would think it should be `{width: 473, height: 355, x: 0, y: 0}`

**Expected behavior**
I would expect that the x value of `croppedAreaPixels` is between 0 and 473 and the y value is between 0 and 355 as we drag the image around the crop area.

trurl-master:
Fixes #240 

### Motivation

- Currently, the `croppedArea` returned by `onCropComplete` when used together with non-zero rotation is not very useful, the values are more like temporary variables necessary for subsequent calculation of `croppedAreaPixels` and do not make much sense on their own. I wanted to change that and unify the output of `croppedArea` and `croppedAreaPixels` when used with rotation and without
- If we want to restore previously stored cropArea we have to use `initialCroppedAreaPixels`. The problem with pixels is that the process that generates them also post-processes them (recalculates based on passed `aspect` and rounds to the nearest integer), which results in slowly drifting crop zone when saved -> restored multiple times

### This solution

- From now on the `croppedArea` return values from 0 to 100 (unless `restrictPosition` is false). It behaves the same as before when rotation is 0˚, but when it's not 0˚ the percentages are calculated within the bounding box size of the rotated image. This approach greatly simplifies crop operations, for `sharp` for example, it's extremely easy to just plug the values into `rotate()`, then `extract()` to get the correct result. (**Breaking change**)
- The `croppedAreaPixels` is also changed, it also returns values within the bounding box of the rotated image with improved accuracy and rounded to the nearest integer (as before) (**Breaking change**)
- Added new prop `initialCroppedAreaPercentages` that uses the exact inverse algorithm that generates `croppedArea`. It minimizes the losses of save/restore operations. They're still present on some occasions as the JS floating-point operations accuracy is limited, but it allows to preserve the crop for much longer. `initialCroppedAreaPixels` still can be used, but it's a bit risky

A PoC of server-side crop using sharp - https://codesandbox.io/s/test-easy-crop-next-e51oi
Official example with client-side crop remade for this PR - https://codesandbox.io/s/react-easy-crop-demo-with-cropped-output-forked-z0cgp

### Migration guide

The main impact this PR has is on the `croppedAreaPixels` when used together with the rotation. As it now generates bounding box relative values, you will need to remake the part that generates the final image, taking that into account.

Here's an example implementation ([codesandbox](https://codesandbox.io/s/react-easy-crop-demo-with-cropped-output-forked-z0cgp)):

First, calculate the bounding box size:
```js
const image = await createImage(...)

const rotRad = getRadianAngle(rotation)
const bBoxWidth =
  Math.abs(Math.cos(rotRad) * image.width) +
  Math.abs(Math.sin(rotRad) * image.height)
const bBoxHeight =
  Math.abs(Math.sin(rotRad) * image.width) +
  Math.abs(Math.cos(rotRad) * image.height)
```

Second, create a canvas element with the size of the bounding box:
```js
const canvas = document.createElement('canvas')

canvas.width = bBoxWidth
canvas.height = bBoxHeight
```

Then, rotate the canvas context around the center of the canvas:
```js
const ctx = canvas.getContext('2d')

ctx.translate(bBoxWidth / 2, bBoxHeight / 2)
ctx.rotate(rotRad)
```

Set context to point to the top-left corner of the rotated image:
```js
ctx.translate(-image.width / 2, -image.height / 2)
```

Draw the image onto the rotated context:
```js
ctx.drawImage(image, 0, 0)
```

What you have in the canvas at this point is a rotated image inside its bounding box. Now, you just need to extract the result using `croppedAreaPixels` and repaint it on a canvas with the size of the final image:
```js
const data = ctx.getImageData(
  croppedAreaPixels.x,
  croppedAreaPixels.y,
  croppedAreaPixels.width,
  croppedAreaPixels.height
)

// set the canvas size to the final image size - this will clear existing context
canvas.width = croppedAreaPixels.width
canvas.height = croppedAreaPixels.height

// paste the extracted image at the top left corner
ctx.putImageData(data, 0, 0)
```

#### `croppedArea`

If you somehow managed to use `croppedArea` with rotation before - you're smarter than me. If not - now is the time. The approach is basically the same as with `croppedAreaPixels` - you just need to convert percentages to pixels based on the image you're cropping

Here's how you can use percentages server-side and crop using sharp ([codesandbox](https://codesandbox.io/s/test-easy-crop-next-e51oi)):

Rotate the image:
```js
const image = sharp(...);

await image.rotate(rotation);
```

Get rotated image bounding box size (basically you need to create a new image out of the rotated one, otherwise the metadata won't change):
```js
const meta = await sharp(await image.toBuffer()).metadata();
```
_Alternatively, you can calculate the bounding box size using the same method we used for the client-side crop_

Convert percentages to pixels:
```js
const cropInfo = {
  top: Math.round((croppedArea.y / 100) * meta.height),
  left: Math.round((croppedArea.x / 100) * meta.width),
  width: Math.round((croppedArea.width / 100) * meta.width),
  height: Math.round((croppedArea.height / 100) * meta.height)
}
```

Extract the image:
```js
const myCroppedImage = await image.extract(cropInfo)
```


ValentinH:
Thanks a lot for this PR and the effort! 🙂

I quickly had a look to the changes from my phone but I'll do a more extensive review of the code this week.

I have a few remarks/suggestions already:
- having to change most e2e tests doesn't sound right, especially for tests not involving rotation. What is the reason of these changes?
- could you please update the cropping logic of the demo website (docs folder) in order to see the actual diff required by these changes? You will need to change the imported library from the latest NPM to a relative path pointing to the src folder.
- in the description of the PR, could you please prepare migration instructions for the breaking changes that we will use when publishing the new major version?

trurl-master:
> Thanks a lot for this PR and the effort! 🙂

> * having to change most e2e tests doesn't sound right, especially for tests not involving rotation. What is the reason of these changes?

The reason for these changes is that to calculate rendered media size, the solution was to use `offsetWidth` and `offsetHeight`. The problem with them is that they report rounded values (contrary to the spec, both firefox and chrome tried to change that but decided not to because it breaks a lot of sites). I can't use `getBoundingClientRect` because it takes transforms into account (unless I render a separate image without rotation just to calculate the width/height).  Instead, I'm calculating the media size manually using the container size (taken with `getBoundingClientRect`) and the fact that the image is placed inside using `max-width: 100%; max-height: 100%`. This move greatly improved the accuracy of crop/zoom restoration. But the side-effect was that the actual CSS values aren't pretty anymore, thus all these changes.

You are correct in saying that it doesn't sound right to change all the e2e tests, but the reason for this in my view is the fact that these tests test implementation details instead of actual results. What matters to the user is what the lib outputs - in our case it's `croppedArea` and `croppedAreaPixels`. This is what should be tested. In this case, the tests wouldn't have been affected (or at least affected much less). Otherwise, every time you make changes to math inside the project, you will end up with "having to change most e2e tests doesn't sound right" moments.

> * could you please update the cropping logic of the demo website (docs folder) in order to see the actual diff required by these changes? You will need to change the imported library from the latest NPM to a relative path pointing to the src folder.

Ah! Sorry, didn't notice it there,. thought it was somewhere else. Updated!

> * in the description of the PR, could you please prepare migration instructions for the breaking changes that we will use when publishing the new major version?

Will do

ValentinH:
I completely agree that current e2e tests are not very valuable. I think that we should replace most of them with visual regression tests as what matters is the visual output.
I will check if the output is actually still the same when testing from the computer.

Regarding using the container size, did you check that it works with all possible values of the `objectFit` property? As the cover ones don't use 100% for max width and height I'm scared it might bring issues.

trurl-master:
Ah! I didn't, that will probably brake, I'll fix it asap. Thank you

trurl-master:
> I completely agree that current e2e tests are not very valuable. I think that we should replace most of them with visual regression tests as what matters is the visual output

Yeah, maybe like Chromatic or something similar, but I think just to test `croppedArea` and `croppedAreaPixels` instead of CSS is a good start

trurl-master:
`objectFit` should be supported properly now

examples/src/index.tsx
```diff
@@ -1,49 +1,134 @@
...
+  constructor(props: {}) {
+    super(props)
+
+    let rotation = 0
+    let initialCroppedAreaPercentages: Area | undefined = undefined
+    let initialCroppedAreaPixels: Area | undefined = undefined
+    let hashType: HashType = 'percent'
+    let imageSrc = imageSrcFromQuery
+
+    if (window && !urlArgs.setInitialCrop) {
```
ValentinH, line 51: I think the hash part was useful while working on this PR but I'd like to keep this example simple. Do you agree to remove the hash-related logic from this file?
However please keep what you added for the image selection and the rotation ticks.

src/index.tsx
```diff
@@ -218,15 +233,64 @@ class Cropper extends React.Component<CropperProps, State> {
...
+                ? {
+                    width: this.containerRect.height * mediaAspect,
+                    height: this.containerRect.height,
+                  }
+                : {
+                    width: this.containerRect.width,
+                    height: this.containerRect.width / mediaAspect,
+                  }
+            break
+          case 'horizontal-cover':
```
ValentinH, line 268: If you try this version on the demo sandbox for `objectFit` (https://codesandbox.io/s/react-easy-crop-forked-1mzi1), something seems wrong for `horizontal-cover` and `vertical-cover`.

Here's the original version: https://codesandbox.io/s/react-easy-crop-forked-p9r34

src/helpers.ts
```diff
@@ -152,46 +175,84 @@ function noOp(_max: number, value: number) {
 }
 
 /**
- * Compute the crop and zoom from the croppedAreaPixels
+ * Compute crop and zoom from the croppedAreaPercentages.
+ */
+export function getInitialCropFromCroppedAreaPercentages(
+  croppedAreaPercentages: Area,
```
ValentinH, line 181: I think that having a single object argument would help to understand what each parameter is when using this function.

src/index.tsx
```diff
@@ -183,29 +186,41 @@ class Cropper extends React.Component<CropperProps, State> {
   }
 
   onMediaLoad = () => {
-    this.computeSizes()
+    const cropSize = this.computeSizes()
     this.emitCropData()
-    this.setInitialCrop()
+    this.setInitialCrop(cropSize as Size)
```
ValentinH, line 191: I don't like to use `as`. Here, in the unlikely case where `cropSize` would be `undefined`, I would prefer to not call `this.setInitialCrop()`.

src/index.tsx
```diff
@@ -218,15 +233,64 @@ class Cropper extends React.Component<CropperProps, State> {
...
+
     if (mediaRef && this.containerRef) {
       this.containerRect = this.containerRef.getBoundingClientRect()
+      const containerAspect = this.containerRect.width / this.containerRect.height
+      const naturalWidth = this.imageRef?.naturalWidth || this.videoRef?.videoWidth || 0
+      const naturalHeight = this.imageRef?.naturalHeight || this.videoRef?.videoHeight || 0
+      const isMediaScaledDown = mediaRef.width < naturalWidth || mediaRef.height < naturalHeight
+      const mediaAspect = naturalWidth / naturalHeight
+
+      // We do not rely on the offsetWidth/offsetHeight if the media is scaled down
```
ValentinH, line 245: I don't know where the issue is but when testing this version with the sandbox showing the "video example", I noticed that the container bounds are not respected anymore when dragging: https://codesandbox.io/s/react-easy-crop-for-videos-forked-w2bl6

image


If you don't manage to fix this, I'm totally fine with keeping the original logic if media is a video. 😉

ValentinH (changes requested):
I've added a few suggestions and reported a few small issues.

Except for this, the PR looks great to me! To be honest, I think this is one of the most advanced and detailed PR I've received on this repository, congrats! 🎉 👏 

Thank you very much for this PR and for adding a detailed migration guide + a nice example with `sharp` (this one should be added to the official list of example once this is merged)! 😍

docs/src/components/Demo/cropImage.ts
```diff
@@ -30,41 +30,43 @@ export default async function getCroppedImg(
...
   }
 
-  const maxSize = Math.max(image.width, image.height)
-  const safeArea = 2 * ((maxSize / 2) * Math.sqrt(2))
+  // calculate bounding box of the rotated image
+  const rotRad = getRadianAngle(rotation)
+  const bBoxWidth =
+    Math.abs(Math.cos(rotRad) * image.width) + Math.abs(Math.sin(rotRad) * image.height)
+  const bBoxHeight =
+    Math.abs(Math.sin(rotRad) * image.width) + Math.abs(Math.cos(rotRad) * image.height)
```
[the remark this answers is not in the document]
trurl-master: I was a little hesitant in making it because there's already a function called `translateSize` that does the same thing but it's much longer than my version, I wasn't exactly sure why, I thought maybe it's more precise. But if it's not, I can replace it with this version to save some space. And maybe rename it to something like `calcBoundingBoxSize`. This way in the codesandbox demos I can just import it and use it instead of writing it manually

docs/src/components/Demo/cropImage.ts
```diff
@@ -30,41 +30,43 @@ export default async function getCroppedImg(
...
   ctx.rotate(getRadianAngle(rotation))
   ctx.scale(flip.horizontal ? -1 : 1, flip.vertical ? -1 : 1)
-  ctx.translate(-safeArea / 2, -safeArea / 2)
+  ctx.translate(-bBoxWidth / 2, -bBoxHeight / 2)
 
-  // draw rotated image and store data.
-  ctx.drawImage(image, safeArea / 2 - image.width * 0.5, safeArea / 2 - image.height * 0.5)
-  const data = ctx.getImageData(0, 0, safeArea, safeArea)
+  // draw rotated image
+  ctx.drawImage(image, bBoxWidth / 2 - image.width * 0.5, bBoxHeight / 2 - image.height * 0.5)
```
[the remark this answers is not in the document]
trurl-master: Ah, good catch, they both work, but the migration guide version is cleaner, so i'll you use it

examples/src/index.tsx
trurl-master: I would argue that it's beneficial to keep the hash part because it helps to test the crop restoration easily - just refresh the page and check that everything stayed the same. For me it's not really an example, more like a development version for people who work on the lib. But up to you, I can remove it if you want

src/index.tsx
trurl-master: You used the older version of the build, before `Support objectFit prop` commit, here's the same codesandbox with the latest version - https://codesandbox.io/s/react-easy-crop-forked-n6kwy

src/helpers.ts
trurl-master: Do you mean to use only one object as a parameter and pass values through this object? I'm not opposed to the idea, but then we need to refactor all the functions in this fashion? I just reused the same coding style the lib uses, for example in `getInitialCropFromCroppedAreaPixels`

ValentinH: Yes this is what I meant. You're right, it would be better to refactor everything at once. We can keep this for later 🙂

src/index.tsx
trurl-master: Should be fixed now - https://codesandbox.io/s/react-easy-crop-for-videos-forked-sljkb

docs/src/components/Demo/cropImage.ts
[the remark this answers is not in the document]
trurl-master: I've updated the demo with the version from the migration guide. Should I change the import in the Demo back to 'react-easy-crop'? Or keep the relative path?

src/index.tsx
trurl-master: Yeah, you're right. Updated

docs/src/components/Demo/cropImage.ts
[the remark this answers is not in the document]
trurl-master: Replace with my version? Do you have any idea why it was created like that?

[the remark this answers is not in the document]
ValentinH: This was contributed by another developer. I'm missing math knowledge to judge. If you have a simpler version that works, let's use it. 🙂

docs/src/components/Demo/cropImage.ts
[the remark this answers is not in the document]
ValentinH: Let's keep it for now so that I can play with it again before approving. I'll try to complete the review tomorrow.

examples/src/index.tsx
ValentinH: Actually while testing in Firefox, I noticed that updating the URL on each event isn't accepted by it and crashes the example page 🙈
image

trurl-master: I'll keep hash updates only for `onCropComplete` then, not for `onCropAreaChange`

ValentinH: Actually, `onCropComplete` is called as well while zooming with the trackpad/mouse or rotating which also triggers the issue. Ideally, we should debounce in the example the logic to update the URL.

Having `onCropComplete` called this way is an issue in my opinion. We might want to debounce it in the future.

trurl-master: Yes, you're right, i've added a debounce with 150ms, that seems to work well

ValentinH (approved):

ValentinH:
I published an alpha version with your branch: `4.0.0-alpha.0`. I'll do a few other tests before merging and releasing 🤞

docs/src/components/Demo/cropImage.ts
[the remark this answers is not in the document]
ValentinH: I updated the helpers to use your technique as it's much shorter.

ValentinH:
This was released as v4.0.0: https://github.com/ricardo-ch/react-easy-crop/releases/tag/v4.0.0

Thank you very much @trurl-master for the awesome work 👏

ValentinH:
@all-contributors please add @trurl-master for code and examples

PR #316, opened 2021-11-29
[CI] Add Cypress cross browser testing

raed667:
Cypress supports running tests on different browsers. I think this will help increase code quality and catch browser-specific bugs.

Documentation: https://docs.cypress.io/guides/guides/cross-browser-testing#Continuous-Integration-Strategies

### Opinion

I have opted for running three different commands on CircleCI, but if you want it to be one command than runs 3 npm scripts, that can also be a solution.


RaedsLab:
Tests are failing because of some rounding issue with Firefox: 

```
Timed out retrying
      + expected - actual

      -'523.5499877929688px'
      +'523.546875px'
```


If you have any ideas, i'm open to fixing it.

ValentinH:
> Tests are failing because of some rounding issue with Firefox:
> 
> ```
>       -'523.5499877929688px'
>       +'523.546875px'
> ```
> 
> If you have any ideas, i'm open to fixing it.

I'm actually wondering how multi-browser tests would work with visual regression tests? 🤔

RaedsLab:
> I'm actually wondering how multi-browser tests would work with visual regression tests? 🤔

If there is a sub-pixel issue we can define some threshold in the comparaison the same way we had to do in the style-guide when we had to deal with shadows being rendered slightly differently between different runs.

ValentinH:
Ok 👌 I think we should implement Percy before trying to fix these browser specific tests issues.

PR #317, opened 2021-11-29
[CI] Split different tests to have better reporting

RaedsLab:
Currently when one step fails on CircleCI the error on Github isn't very obvious.

This PR splits the different testing into different jobs.

image

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,28 +1,6 @@
 version: 2.1
 
-aliases:
-  - &restore_yarn_cache
-    restore_cache:
-      name: Restore yarn cache
-      keys:
-        - v1-yarn-{{ arch }}-{{ .Branch }}-{{ checksum "yarn.lock" }}
-        - v1-yarn-{{ arch }}-{{ .Branch }}
-        - v1-yarn-
-  - &restore_node_modules
-    restore_cache:
-      name: Restore node_modules cache
-      keys:
-        - v1-node-modules-{{ arch }}-{{ .Branch }}-{{ checksum "yarn.lock" }}
-        - v1-node-modules-{{ arch }}-{{ .Branch }}
-        - v1-node-modules-
-
 executors:
-  node:
-    docker:
-      - image: cimg/node:16.13.0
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -1,34 +1,114 @@
...
-    - run:   # run coveralls
-        name: Run e2e tests
-        command: |
-          yarn e2e:ci
+      - checkout
+      - attach_workspace:
+          at: .
+      - run:
+          name: Installing Cypress Binary
+          command: yarn cypress install
```
RaedsLab, line 111: This was needed because of the following error: 

```
The cypress npm package is installed, but the Cypress binary is missing.

We expected the binary to be installed here: /root/.cache/Cypress/5.2.0/Cypress/Cypress

Reasons it may be missing:

- You're caching 'node_modules' but are not caching this path: /root/.cache/Cypress
- You ran 'npm install' at an earlier build step but did not persist: /root/.cache/Cypress

Properly caching the binary will fix this error and avoid downloading and unzipping Cypress.

Alternatively, you can run 'cypress install' to download the binary again.

https://on.cypress.io/not-installed-ci-error
```

If you have time, we can also try caching this during `setup` and avoid double installing:  
image

ValentinH: OK now I know where I read that (when we discussed about it at Ricardo and caching node_modules was giving me bad vibes) ^^

I guess we should listen to: 
image

RaedsLab: This is only the case because cypress has a binary that it download outside node_modules. We can cache both as far as i understand.

ValentinH: Yes we can try this

RaedsLab: When you have some time you can have a look. I think this solution is aligned with their recommendations 

https://glebbahmutov.com/blog/do-not-let-cypress-cache-snowball/

ValentinH:
Thanks for this!

ValentinH:
@all-contributors please add @raed667 for infra

PR #323, opened 2021-12-03
Fix Cropper defaultProps types

cvolant:
## Motivation

```js
import dynamic from 'next/dynamic'
const Cropper = dynamic(() => import('react-easy-crop'))
```
Using the above [Next.js](https://nextjs.org/) dynamic import, Typescript complains that Cropper defaultProps are not assignable to Cropper props.
Why? Because defaultProps set `cropShape` to `'rect'` and `objectFit` to `'contain'`, which is good since:
```ts
export type CropperProps = {
  // ...
  cropShape: 'rect' | 'round'
  objectFit?: 'contain' | 'horizontal-cover' | 'vertical-cover'
  // ...
}
```
... but doing that, TS consider that defaultProps set `cropShape` and `objectFit` to a `string`.

## What this PR does

It merely adds an `as const` behind the values given to `cropShape` and `objectFit` in defaultProps.


raed667:
Would using TypeScript [Enums](https://www.typescriptlang.org/docs/handbook/enums.html) also solve the issue ?

ValentinH:
> Would using TypeScript [Enums](https://www.typescriptlang.org/docs/handbook/enums.html) also solve the issue ?

I'm not a big fan of enums in general as they don't bring anything more than literals unions and are not supported by Babel.

I'm not sure but if we use enums for the props definition, library users will have to use the enum as well when setting the props.

I'll merge and release this PR once I'm on the computer. I also want to check what the d.ts output looks like.

ValentinH:
All good for me 🙂

ValentinH:
Released as [v4.0.1](https://github.com/ricardo-ch/react-easy-crop/releases/tag/v4.0.1). Thank you for your contribution 🎉

ValentinH:
@all-contributors please add @cvolant for code

PR #336, opened 2021-12-29
add alternative way using zoom without material ui slider for codesandbox example

oleksi-l:
I dont want add to my project excessive dependencies, so I write my own version for documentation without material ui slider. I tried to reproduce all functionality of material ui slider, but styling effects little bit another, please check and if this on your opinion will be useful for project - merge to library.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ A React component to crop images/videos with easy interactions
 
 Check out the examples:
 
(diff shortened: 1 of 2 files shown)


ValentinH:
Thanks for this suggestion. I decided to directly edit the original sandbox with your changes so that I stay the owner of it for security reason.

ValentinH:
@all-contributors  please add @oleksi-l for documentation

PR #357, opened 2022-02-22
Add requireCtrlKey & requireMultiTouch props

MatixYo:
https://github.com/ValentinH/react-easy-crop/issues/356


ValentinH (commented):
Nicely done 🙂

I think we should provide a way to enable the parent to show a message when the user tries to interact with the cropper but the props are blocking it. The goal is to be able the same kind of UI than Maps.
One callback for each prop should do the job, like `onInteractionBlockedByCtrlKey()` (or something else if you find a better naming 😅).

MatixYo:
This could be achieved by listening for onTouchMove & onWheel event on the parent so I'm not really sure if it's needed 😅

ValentinH:
I gave this a second thought and I think I found a way that combines your approach and my requirements of providing an event to build custom UI upon.

Instead of using boolean options, I would use directly 2 callbacks:
```tsx
  onDragRequest: (
    e: React.MouseEvent<HTMLDivElement, MouseEvent> | React.TouchEvent<HTMLDivElement>
  ) => boolean
  onWheelRequest: (e: WheelEvent) => boolean
```

The first one being called on mouse down:
```tsx
  onMouseDown = (e: React.MouseEvent<HTMLDivElement, MouseEvent>) => {
    if (!this.props.onDragRequest(e)) {
      return
    }
    e.preventDefault()
    document.addEventListener('mousemove', this.onMouseMove)
    document.addEventListener('mouseup', this.onDragStopped)
    this.onDragStart(Cropper.getMousePoint(e))
  }
```
and on touch start:
```tsx
  onTouchStart = (e: React.TouchEvent<HTMLDivElement>) => {
    if (e.touches.length === 2) {
      this.onPinchStart(e)
    } else if (e.touches.length === 1) {
      if (!this.props.onDragRequest(e)) {
        return
      }
      this.onDragStart(Cropper.getTouchPoint(e.touches[0]))
    }
    document.addEventListener('touchmove', this.onTouchMove, { passive: false }) // iOS 11 now defaults to passive: true
    document.addEventListener('touchend', this.onDragStopped)
  }
```

and the second one on wheel:
```tsx
  onWheel = (e: WheelEvent) => {
    if (!this.props.onWheelRequest(e)) {
      return
    }

    e.preventDefault()
   // rest of the logic
  }
```

Here how it would be used:
image

What do you think about this approach?

MatixYo:
If you want it to be more flexible then I'd move onDragRequest higher inside onTouchStart so that you can prevent also in other scenarios. My approach was more strict but I wanted to make it work as on Maps.

```
  onTouchStart = (e: React.TouchEvent<HTMLDivElement>) => {
    if (this.props.onDragRequest && !this.props.onDragRequest(e)) {
      return
    }
    if (e.touches.length === 2) {
      this.onPinchStart(e)
    } else if (e.touches.length === 1) {
      this.onDragStart(Cropper.getTouchPoint(e.touches[0]))
    }
    ...
  }
```

ValentinH:
I get your point but then the name doesn't really match anymore : `onDragRequest` while it also called for pinch gestures. 🤔 

What do you think?

MatixYo:
So maybe just change name to onTouchRequest?

ValentinH:
Yes , and we can move the part in onMouseDown for now as we don't need it for the original issue this PR is solving.

Would you like to do the changes or should I directly do it?

ValentinH:
I created this demo: https://codesandbox.io/s/react-easy-crop-scroll-demo-fp21qd

The new feature from the PR is available in v4.1.0 🚀 

Thanks @MatixYo for the idea and the contribution 👍

ValentinH:
@all-contributors please add @MatixYo  for code

ValentinH:
@all-contributors please add @MatixYo  for code

PR #365, opened 2022-03-05
docs: add npm example in development section

raed667:
Add example of how to run development with `npm`

Also, make `yarn` example one command line instead of 2


ValentinH:
Thanks for the test!

PR #366, opened 2022-03-20
[CI] Add Quality Gate Action

raed667:
Add a "quality-gate" GitHub Action that runs `build`, `lint` then `unit test`.
The action can run on push, PR, or manually. 

I added a badge to the read-me but I can't see if it will properly work until it runs on the main repo.

Additional changes: I had to add a condition to the release workflow so that the it wouldn't fail on my fork. 

Next steps:

1. Remove the duplicated workflows from CircleCI
2. Create a Github workflow for Cypress 
3. Add better integration for test results, linting, coverage reporting, etc .. (maybe something like https://danger.systems/js/ )

### DIFF
--- a/.github/workflows/quality-gate.yml
+++ b/.github/workflows/quality-gate.yml
@@ -26,6 +26,12 @@ jobs:
 
       - name: Create Build
         run: yarn build
+      - name: Archive build artifacts
+        uses: actions/upload-artifact@v3
+        with:
+          name: dist
+          path: |
+            dist
 
       - name: Run linting
         run: yarn lint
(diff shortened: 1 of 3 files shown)


.github/workflows/quality-gate.yml
```diff
@@ -0,0 +1,40 @@
+name: Quality Gate
+
+on: [push, pull_request, workflow_dispatch]
```
ValentinH, line 3: What is `workflow_dispatch` for?

raed667: To manually trigger the action if needed. 

https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/

ValentinH:
Thank for this!

PR #368, opened 2022-03-20
[CI] Add Cypress tests to Github Actions

raed667:
Adds Cypress tests to GitHub Actions.

A few changes are required for this to work correctly: 

Go to project [secrets](https://github.com/raed667/react-easy-crop/settings/secrets/actions) and add the following secrets: 

1- CYPRESS_RECORD_KEY (Available in dashboard.cypress.io)
2- CYPRESS_PROJECT_ID (`ie9fhh` since its easier to manage if it is not hard-coded)
3- PERCY_TOKEN

(you'll need to setup `CYPRESS_PROJECT_ID` in circleci as well if you want the tests to run in both platforms for the moment)


package.json
```diff
@@ -92,7 +92,7 @@
     "np": "^6.5.0",
     "prettier": "^2.1.2",
     "query-string": "^6.1.0",
-    "raw-loader": "^4.0.1",
+    "raw-loader": "^4.0.2",
```
raed667, line 95: Just updated this so i can do some cache busting

.github/workflows/cypress.yml
```diff
@@ -0,0 +1,36 @@
...
+          restore-keys: |
+            yarn-deps-${{ hashFiles('yarn.lock') }}
+      - run: yarn install --frozen-lockfile
+
+      - name: Run e2e tests
+        uses: cypress-io/github-action@v2
+        with:
+          build: yarn build
+          start: yarn e2e:ci
+        env:
```
ValentinH, line 32: On the CircleCI job, there is a `CYPRESS_CACHE_FOLDER` env, don't we need it here as well?

.github/workflows/cypress.yml
```diff
@@ -0,0 +1,36 @@
+name: E2E Tests
+
+on: [push]
```
ValentinH: also, why not doing it on pull requests as well ?

ValentinH:
Once this is merged, we will shutdown the CircleCI integration so no need to update the env variables there.

.github/workflows/cypress.yml
raed667: Yes we probably should !

.github/workflows/cypress.yml
raed667: Yes I remember something about caching the install of Cypress itself (besides caching the usual node_modules), so far it didn't cause an issue. But I'll take a bit of time to read on it to be sure if it is needed in this case or not.

.github/workflows/cypress.yml
ValentinH: Ok, let's do it then 🙂

.github/workflows/cypress.yml
raed667: So if i understand correctly `cypress-io/github-action@v2` already does it. 

However, if we want to cache node_modules independently, we can use the following example https://github.com/bahmutov/cypress-gh-action-split-install/blob/master/.github/workflows/tests.yml 

so unless we want to do advanced parallel runs and caching, it should be fine as is.

ValentinH:
Tests are failing because env variables are not available. I think this is probably because by default GitHub doesn't expose env variables to PR coming from forks... So annoying.
I will check this but I had a similar issue for the release workflow.

raed667:
I was not aware of such limitations, the PR is small enough, you can duplicate it on your end if you'd like.

ValentinH:
> I was not aware of such limitations, the PR is small enough, you can duplicate it on your end if you'd like.

The problem is that this limitation will happen for every PR which goes against the goal of running E2E tests to check that PRs are not breaking something 🙈 .
image

I don't really know how to solve this and couldn't find anything from Percy side.

raed667:
It seems like quite a big limitation. I think losing the Cypress video recordings is alright, but Percy is too useful. 

Then lets just keep it on CircleCI ..

ValentinH:
Yes I agree unfortunately 😞

raed667:
Closing !

PR #379, opened 2022-05-02
feature: add auto-cover

### LINKED ISSUE #378: Add objectFit: "auto-cover"
Some people already started discussing this at #186 

The idea would be to create a new objectFit value which would automatically find the best fit based on the images width/height.

This can be done in user land like so:
```
... calculated the width and height of the image ...

<Cropper
  ...
  objectFit={width > height ? 'vertical-cover' : 'horizontal-cover'}
/>
```
 but it would be nice if the package itself contined it as it is a fairly common use case.

vassbence:
closes #378


vassbence:
@ValentinH let me know if this is how it's supposed to be done :)

ValentinH:
Yes I think it's nice like this! Could you please add this new option to the Readme please and explain its purpose?

ValentinH (approved):

ValentinH:
Thank you for this! This will be automatically published in a few minutes 🎉

ValentinH:
@all-contributors please add @vassbence for code contribution

PR #384, opened 2022-06-02
Add ability to play/pause and mute/unmute video using props

AnthonyUtt:
Needed for a current application, figured I would add a PR upstream in case anyone else wants this sort of functionality.

Both props are completely optional, omitting both will not impose any breaking changes to existing cropper setups.


AnthonyUtt:
Here's a working CodeSandbox example: https://codesandbox.io/s/react-easy-crop-forked-0mf3nf

ValentinH:
Thanks for this PR!

I'm wondering if we could actually make this more generic and cover more use cases by directly exposing the `video` `ref` (same for the image). What do you think?

AnthonyUtt:
That's a great idea! Let me jam on it a bit and I will update shortly.

AnthonyUtt:
@ValentinH I have updated with optional callback props to set both the image and video refs in the parent component. Theoretically, this could be updated to use a single `mediaRef` object, but that implementation would require a bit more of a refactor of the existing component so I opted for the simpler solution. Please let me know if there are any issues.

Edit: CodeSandbox working example: https://codesandbox.io/s/react-easy-crop-forked-6rtj6b?file=/src/index.js

ValentinH:
Thanks for the contribution!

Issue #139, opened 2020-05-22
Inline style and CSP policy


sasha-3ap:
When CSP security policy with directive style-src is set, unless 'unsafe-inline' value is used, react-easy-crop inline style are ignored which results in incorrect react-easy-crop render. As setting 'unsafe-inline' defeats purpose of CSP, react-easy-crop should provide solution without inline styles. Please check '[Disable inline style](https://dontkry.com/posts/code/disable-inline-styles.html)' from more information and possible solution.

If I solve this I will create PR.

ValentinH:
You are already able to request the CSS explicitly: see https://github.com/ricardo-ch/react-easy-crop/blob/master/README.md#styles.

Does it cover your use case?

sasha-3ap:
Seems it does not. Even if we use disableAutomaticStylesInjection and css it is still [added as inline style](https://github.com/ricardo-ch/react-easy-crop/blob/master/src/index.tsx#L106). In render we have inline style [here](https://github.com/ricardo-ch/react-easy-crop/blob/master/src/index.tsx#L467), [here](https://github.com/ricardo-ch/react-easy-crop/blob/master/src/index.tsx#L484) and  [here](https://github.com/ricardo-ch/react-easy-crop/blob/master/src/index.tsx#L494). 

In Material UI CSP with nonce is [supported](https://material-ui.com/styles/advanced/#how-does-one-implement-csp) for inline styles by reading nonce from meta tag. Similar solution could be used in react-easy-crop, add nonce from meta tag. Or add style through javascript. Or ...

ValentinH:
I understand for the inline styles in render. However, the styles injected in the `head` are not if you set `disableAutomaticStylesInjection` to true.

Just to understand, is this issue a regression of version 3.0?

sasha-3ap:
I tried adding disableAutomaticStylesInjection to Cropper and import css in component that used it. Is this a proper way? I would still error:
![image](https://user-images.githubusercontent.com/54268606/82900368-b58dd400-9f5c-11ea-86ba-20e2ac325599.png)


Version is latest. Not sure if this can be considered as regression.  

We generate nonce on every new application request but we also insert nonce in meta tag (as required by MUI) so it can be used in inline styles.

ValentinH:
What's the `injectStylesIntoStyleTag` function?

sasha-3ap:
[...]

PR #387, opened 2022-06-14
Add ability to provide a nonce for style injection

seanparmelee:
Hello! 👋 
This is somewhat related to #139 in that we're also using a content security policy that's preventing the inline styles from being applied.

I realize that one way around this is to pass the `disableAutomaticStylesInjection` prop and add the styles manually, but I was curious if you'd be open to adding a `nonce` prop so that the injected `style` tag can have its `nonce` attribute set and we can continue to use automatic style injection along with our CSP.


ValentinH:
Thanks for this contribution. I understand that this could be useful for some people and avoid having to opt-out of the automatic style injection. Let's merge this!

PR #404, opened 2022-08-11
Expose `getInitialCropFromCroppedArea[Percentages|Pixels]` functions

AnthonyUtt:
When working with the cropper, it is helpful to be able to align the media with a specific edge of the crop box. Forwarding these two functions to the user would allow one to simply derive the intended `croppedAreaPercentages` or `croppedAreaPixels` and pass them on to their respective methods to get the required `crop` and `zoom` values. I didn't want to use the `initialCroppedAreaPixels/Percentages` because it is currently only called when the assets are loaded and would require unloading/reloading the assets.

Working example on CodeSandbox: https://codesandbox.io/s/react-easy-crop-forked-3v0hi3

Definitely open to suggestions if passing the `mediaSize` and `cropSize` values up to the parent is frowned-upon.


AnthonyUtt:
Not sure why the extra commits are showing as those should have already been merged. I think it has to do with how I merged upstream changes.

ValentinH:
Thanks for this PR!
This is definitely intended for advanced usage but I think this is a valid use case with little changes on the API.

Could you please add the 2 new props to the Readme with maybe a mention to your codesandbox example?

AnthonyUtt:
@ValentinH Absolutely. I will push an update shortly.

AnthonyUtt:
@ValentinH I believe this should be ready to merge. Let me know if you need anything else from me.

ValentinH (approved):
Thank you!

PR #410, opened 2022-09-22
Add support for running an srcDoc style iFrame

glendaviesnz:
As [noted here](https://github.com/WordPress/gutenberg/issues/38557), there are some instances where iFraming `react-easy-crop` causes the styles and events to be lost as they are always attached to the global `window` and `document`.

This PR changes the `Cropper` `componentDidMount` method to instead get the `document` and `window` references from the current `containerRef` if it is defined. 

The PR also adds an iFramed version of the component to the example page to allow for testing. This can be accessed via `http://localhost:3001/?iframed=true`.

### DIFF
--- a/examples/src/iframe.tsx
+++ b/examples/src/iframe.tsx
@@ -39,6 +39,7 @@ export default function Iframe({ children }: Props) {
         ref={iFrameRef}
         srcDoc="<!doctype html>"
         title="test iframed"
+        data-cy="iframe"
       >
         <>{iframeBody && createPortal(children, iframeBody)}</>
       </iframe>
--- a/examples/src/iframe.tsx
+++ b/examples/src/iframe.tsx
@@ -1,0 +1,43 @@
+// @ts-nocheck
+import * as React from 'react'
+
+import './styles.css'
+import { createPortal } from 'react-dom'
+
+export default function Iframe({ children, ...props }) {
+  const [iframeBody, setIframeBody] = React.useState<HTMLElement>()
+
+  const iFrameRef = React.useRef<HTMLIFrameElement>()
+
+  React.useEffect(() => {
+    function setDocumentIfReady() {
+      const { contentDocument } = iFrameRef.current
+      const { readyState, documentElement } = contentDocument
+
+      if (readyState !== 'interactive' && readyState !== 'complete') {
+        return false
+      }
+
+      setIframeBody(documentElement.getElementsByTagName('body')[0])
+
+      return true
+    }
+
+    // Document set with srcDoc is not immediately ready.
+    iFrameRef.current.addEventListener('load', setDocumentIfReady)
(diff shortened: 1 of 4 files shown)


examples/src/iframe.tsx
```diff
@@ -0,0 +1,43 @@
+// @ts-nocheck
```
glendaviesnz: I ran out of time to sort some typing issues in this file, will come back to it and fix if general approach is approved.

src/Cropper.tsx
```diff
@@ -103,29 +103,37 @@ class Cropper extends React.Component<CropperProps, State> {
...
 
   state: State = {
     cropSize: null,
     hasWheelJustStarted: false,
   }
 
   componentDidMount() {
-    window.addEventListener('resize', this.computeSizes)
     if (this.containerRef) {
+      if (this.containerRef.ownerDocument) {
```
glendaviesnz, line 116: I did also try adding some extra props to allow passing in of a window and document ref, but it seemed less efficient if using the likes of a `useRef` in the parent component to get this, as it took a couple of renders to get the `ref.current` set and then pass these down, so setting them in this component, pushing styles into doc, etc. then needed to be handled in `componentDidUpdate`.

glendaviesnz:
@ValentinH have just left this as draft for now so you can check if you are happy with the general approach or not - if you are I will sort the remaining type issues and add some extra tests before getting a full review.

src/Cropper.tsx
ValentinH: If this approach fixed all the issue on Gutenberg side, it's great as it doesn't grow the component external API.

Just by curiosity, why trying to pass window and document refs instead of directly passing the window and document instances?

glendaviesnz: > why trying to pass window and document refs instead of directly passing the window and document instances

we sort of have the same problem on the Gutenberg end in that within the component where `react-easy-crop` is used the default window and document references are the globals, so to get the iframe ones we have to work up from the current component with the same sort of `componenrRef.ownerDocument` approach - does that make sense?

ValentinH: Yes I get it 👍

glendaviesnz: If you are happy with this approach then I will get it tidied up and submitted for review tomorrow

ValentinH: Alright, looking forward to it. 🎉

glendaviesnz:
@ValentinH I think it should be good to go. All the existing e2e tests pass with the change in place, and I added one additional test to make sure the styles get applied to the iframe doc - let me know if you think we need any additional tests on top of this.

ValentinH:
All good!

glendaviesnz:
Big thanks for the quick turnaround on this @ValentinH!

PR #425, opened 2022-12-02
Add package.json css export

buhavon:
# Purpose of the Pull Request
## Problem
Using vite bundler in a monorepo setup with the `disableAutomaticStylesInjection`  set to true, I was facing an issue where I couldn't import the `react-easy-crop.css` using a reference to the package, like `import 'react-easy-crop/react-easy-crop.css'`. Instead I had to import this file from the node_modules, which is not ideal and actually break if the node_modules location change.

## Solution
Adding an export statement in the package.json to be able to import the `react-easy-crop.css` by referencing the package name.

### DIFF
--- a/scripts/copy-build-files.js
+++ b/scripts/copy-build-files.js
@@ -31,8 +31,8 @@ async function createPackageFile() {
         types: './index.d.ts',
       },
       './react-easy-crop.css': {
-        'import': './react-easy-crop.css',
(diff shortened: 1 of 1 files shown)


ValentinH:
Thank you for the fix and the detailed explanation.

buhavon:
Great ! Thanks for the quick merge.

PR #438, opened 2023-01-26
Mark props with default values as optional

intrnl:
This pull requests marks props that has default values as optional

### DIFF
--- a/src/Cropper.tsx
+++ b/src/Cropper.tsx
@@ -20,16 +20,16 @@ export type CropperProps = {
   video?: string | VideoSrc[]
   transform?: string
   crop: Point
-  zoom: number
-  rotation: number
-  aspect: number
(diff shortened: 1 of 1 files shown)


ValentinH:
I need to try it locally (Circle CI was disconnected since their hack). Indeed, this used to cause issues with the defaultProps definition and usage in the component implementation. 

Did you try running type checking on this?

ValentinH:
I checked and it is what I fear: if you mark them as optional then they are seen as optional in the implementation despite of the default props:
image


What issue are you facing because these props are not required when using the component:
image
(there is no error above)

intrnl:
mmm yeah that's unfortunate, I hadn't actually checked the typings properly because I didn't have the time to set the project up locally. I think I'll close this for now.

PR #472, opened 2023-06-20
Switch putImageData method to drawImage

carlosdi0:
Hi, this change fix this issue [https://github.com/ValentinH/react-easy-crop/issues/470](https://github.com/ValentinH/react-easy-crop/issues/470).

[Codesandbox fixed](https://codesandbox.io/s/react-easy-crop-custom-image-demo-forked-dxwq7g?file=/src/canvasUtils.js)

### DIFF
--- a/docs/src/components/Demo/cropImage.ts
+++ b/docs/src/components/Demo/cropImage.ts
@@ -45,23 +45,37 @@ export default async function getCroppedImg(
   // draw rotated image
   ctx.drawImage(image, 0, 0)
 
-  // croppedAreaPixels values are bounding box relative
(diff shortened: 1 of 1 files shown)


ValentinH:
Thank you for this. So if I understand correctly, we now need to use 2 canvas for this to work?

carlosdi0:
Yes, just that and besides, from what I've read drawImage is much more optimal than putImageData.

[Benchmark](https://www.measurethat.net/Benchmarks/Show/9510/0/putimagedata-vs-drawimage)

ValentinH:
Awesome thanks! I'll update the pinned demo from the readme then

PR #484, opened 2023-09-07
fix: consider page scroll when retrieving container point

pontusdacke:
Suggested solution for https://github.com/ValentinH/react-easy-crop/issues/482

Takes page scroll in account in getContainerPoint so zoom with mousewheel works correct when restrictPosition is set to false.

### DIFF
--- a/src/Cropper.tsx
+++ b/src/Cropper.tsx
@@ -116,6 +116,7 @@ class Cropper extends React.Component<CropperProps, State> {
   currentDoc: Document | null = typeof document !== 'undefined' ? document : null
   currentWindow: Window | null = typeof window !== 'undefined' ? window : null
   resizeObserver: ResizeObserver | null = null
+  initialScroll: { scrollX: number, scrollY: number} | null = null
 
   state: State = {
     cropSize: null,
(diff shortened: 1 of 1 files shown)


ValentinH:
Thanks for this PR. I'll have a look and play a bit with this solution. I reminds me of an issue I had to address on https://github.com/ValentinH/react-easy-sort where I had to use some properties that were already taking the scroll into account.

In the end, the issue isn't related with `restrictPosition=false`, it's also present when it's set to `true` but it's less obvious.

ValentinH:
I checked what was done on `react-easy-sort` and basically, we are storing the container position in the window on every gesture start and then we use this position to translate the event point (which is relative to the window) to the position inside the container.

Relevant piece of code:
- [reading the position](https://github.com/ValentinH/react-easy-sort/blob/b3b3b9e054eb823ef1bdae3783d21ebe55f344a3/src/hooks.ts#L100-L101)
- [saving the container position](https://github.com/ValentinH/react-easy-sort/blob/b3b3b9e054eb823ef1bdae3783d21ebe55f344a3/src/hooks.ts#L164)

pontusdacke:
I applied your suggestion, storing the position on each gesture. It seem to have achieved the same results, fixing our reported issue.

ValentinH (approved):

ValentinH:
Thank you for the quick fix! Merging this, it will be automatically published in a few minutes 🎉

PR #506, opened 2023-12-14
fix(crop-area): recalculate sizes on media objectFit change

kruchkou:
Hello @ValentinH!

I've faced an issue with crop area size computation on picture changes with obejct-fit: "cover".

## Pre requirements
- Use round cropper with fixed size 

Cropper properties
```
object-fit: "cover",
aspect: 1,
cropShape: "round",
```
Container css properties:
```
width: 500px,
height: 500px
```

## Steps to reproduce:
- Choose Horizontal Big image (all sides > 500px)
- Change image to Vertical Small (all sides < 500px)

(Same for Vertical Big - Horizontal Small image pair)

## Expected behaviour

Crop area is 500x500
Screenshot 2023-12-14 at 15 04 12

## Observed behaviour

Crop area is 250x250
Screenshot 2023-12-14 at 14 55 36

To fix the issue I've moved the getObjectFIt() result value into state and added callback to recalculateSize() on this state value change.

Issue code example: https://codesandbox.io/p/sandbox/react-easy-crop-demo-forked-6xh2m2?file=%2Fsrc%2Findex.js
Fix code example: https://codesandbox.io/p/sandbox/react-easy-crop-forked-vzrfyk?file=%2Fsrc%2Findex.js

### DIFF
--- a/src/Cropper.tsx
+++ b/src/Cropper.tsx
@@ -67,6 +67,7 @@ export type CropperProps = {
 type State = {
   cropSize: Size | null
   hasWheelJustStarted: boolean
+  mediaObjectFit: String | undefined
 }
 
 const MIN_ZOOM = 1
@@ -121,6 +122,7 @@ class Cropper extends React.Component<CropperProps, State> {
   state: State = {
     cropSize: null,
     hasWheelJustStarted: false,
+    mediaObjectFit: undefined
   }
 
   componentDidMount() {
@@ -217,6 +219,11 @@ class Cropper extends React.Component<CropperProps, State> {
     }
     if (prevProps.video !== this.props.video) {
       this.videoRef.current?.load()
+    }
+
+    const objectFit = this.getObjectFit();
+    if (objectFit !== this.state.mediaObjectFit) {
+      this.setState({mediaObjectFit: objectFit}, this.computeSizes)
     }
   }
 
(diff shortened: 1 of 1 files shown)


src/Cropper.tsx
```diff
@@ -218,6 +220,11 @@ class Cropper extends React.Component<CropperProps, State> {
     if (prevProps.video !== this.props.video) {
       this.videoRef.current?.load()
     }
+
+    const objectFit = this.getObjectFit();
```
ValentinH, line 224: Do we need to store this in the state? Couldn't just call `this.computeSizes` when this prop changes (like done for other props above)?

ValentinH (commented):
Thank you for catching this and providing a fix. I have a question though before approving this PR.

src/Cropper.tsx
kruchkou: We could not catch this through props since it's not the _props.objectFit_ changes

What is being changed is the result of _getObjectFit()_ method that calculates how to properly align the image inside the container (should horizontal-cover or vertical-cover be applied)

- To determine which one should be applied the image should already be presented on the layout to get its sizes (that's why requires re-render)
- Since we already calculated the sizes and props haven't changed on the second render run, the componentDidUpdate() doesn't call _computeSIzes()_ method

Thats why on _getObjectFit()_ result change we need a callback for _computeSizes()_ and re-render (on state change) to apply new sizes

ValentinH:
Thanks again for this PR!

PR #527, opened 2024-04-03
chore: improve package exports

### LINKED ISSUE #490: Cannot use import statement outside a module
**Describe the bug**
I'm getting `Cannot use import statement outside a module` when building it for prod, dev just works fine. I've `type: module` in my package.json and here's my tsconfig
```json
"esModuleInterop": true,
"module": "NodeNext",
"moduleResolution": "NodeNext",
"target": "ES2022",
```
And when importing `Cropper` like this `import Cropper from 'react-easy-crop'` gives this error when using it
```
JSX element type 'Cropper' does not have any construct or call signatures.ts(2604)
```
This project did build successfully before with these exact configs but after adding `react-easy-crop`, it doesn't

Full error:
```console
(node:406) Warning: To load an ES module, set "type": "module" in the package.json or use the .mjs extension.
(Use `node --trace-warnings ...` to show where the warning was created)
SyntaxError: Cannot use import statement outside a module
/myapp/node_modules/react-easy-crop/index.module.js:1
import { __assign, __extends } from 'tslib';
^^^^^^
SyntaxError: Cannot use import statement outside a module
```

**Expected behavior**
It should build successfully with these configs and shouldn't throw an error when using the `Cropper` component

ValentinH:
This is an attempt to fix #490.

However, https://arethetypeswrong.github.io/ is still not fully happy about it. The problem is that we are using `bili` to build the library and it's not maintain anymore. Ideally, we need to migrate to a fully new setup to build this lib but I'm a bit lost at the moment (and I hate this topic!).

<details>
  <summary>📦 Published PR as canary version: <code>5.0.7--canary.527.1aad632.0</code></summary>
  <br />

  :sparkles: Test out this PR locally via:
  
  ```bash
  npm install react-easy-crop@5.0.7--canary.527.1aad632.0
  # or 
  yarn add react-easy-crop@5.0.7--canary.527.1aad632.0
  ```
</details>

### DIFF
--- a/bili.config.ts
+++ b/bili.config.ts
@@ -23,7 +23,7 @@ const config: Config = {
       config.output.fileName = 'umd/react-easy-crop[min].js'
     }
     if (format === 'esm') {
-      config.output.fileName = '[name].module.js'
+      config.output.fileName = '[name].module.mjs'
     }
     return config
   },
--- a/scripts/copy-build-files.js
+++ b/scripts/copy-build-files.js
@@ -22,7 +22,7 @@ async function createPackageFile() {
     exports: {
       '.': {
         import: {
-          types: './index.d.ts',
+          types: './index.d.mts',
           default: './index.module.mjs',
         },
         require: {
--- a/scripts/copy-build-files.js
+++ b/scripts/copy-build-files.js
@@ -22,7 +22,7 @@ async function createPackageFile() {
     exports: {
       '.': {
         import: {
-          types: './index.d.mts',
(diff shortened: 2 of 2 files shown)


ValentinH:
Hint: try https://github.com/egoist/tsup and follow https://www.totaltypescript.com/how-to-create-an-npm-package

However, IIRC, last time I tried `tsup`, I had issues with inlining the CSS into the lib.

ValentinH:
Better fix landed on main

PR #551, opened 2024-07-25
fix(crop-area): initial render with defined objectFit for initialCrop…

### LINKED ISSUE #550: InitialCroppedAreaPixels no longer works properly after v5 upgrade
**Describe the bug**
After upgrading to react-easy-crop 5 I noticed a few things:
1) The image is first loaded and rendered on the screen without any max-width/max-height. This flashes the image for 1 frame on the browser. 
2) The `onMediaLoaded` callback also reflects 1) as the returned media size has the same natural and current width/height.
3) The `initialCroppedAreaPixels` no longer works properly because it uses the wrong media size for calculation
4) The `setMediaSize` callback is called twice on component mount. The second time it's called with the correct media size but it's too late for the correct `initialCroppedAreaPixels` to work.

I suspect that this PR broke the initialCroppedAreaPixels feature: [#506](https://github.com/ValentinH/react-easy-crop/pull/506) because it doesn't set a default mediaObjectFit inside the Cropper.tsx fast enough. It does it after the initial render. The initial render is completed with this value being undefined.

**Expected behavior**
I expected the prop `initialCroppedAreaPixels` to render the cropper with the state of the saved crop.

itsUndefined:
Fixes: [#550](https://github.com/ValentinH/react-easy-crop/issues/550)

### DIFF
--- a/src/Cropper.tsx
+++ b/src/Cropper.tsx
@@ -742,7 +742,7 @@ class Cropper extends React.Component<CropperProps, State> {
       classes: { containerClassName, cropAreaClassName, mediaClassName },
     } = this.props
 
-    const objectFit = this.state.mediaObjectFit
+    const objectFit = this.state.mediaObjectFit ?? this.getObjectFit()
 
     return (
       <div


itsUndefined:
I also confirmed that this change doesn't affect the fix for #506 so it's safe

src/Cropper.tsx
```diff
@@ -742,7 +742,7 @@ class Cropper extends React.Component<CropperProps, State> {
       classes: { containerClassName, cropAreaClassName, mediaClassName },
     } = this.props
 
-    const objectFit = this.state.mediaObjectFit
+    const objectFit = this.state.mediaObjectFit ?? this.getObjectFit()
```
ValentinH, line 745: Could we instead add the following code inside the `componentDidMount()`:
```ts
const objectFit = this.getObjectFit()
if (objectFit !== this.state.mediaObjectFit) {
  this.setState({ mediaObjectFit: objectFit }, this.computeSizes)
}
```

I think it would fit better with the current logic of having it stored in the state.

ValentinH (commented):
Thank you for reporting the issue and providing a fix. I have one suggestion though 🙂

src/Cropper.tsx
itsUndefined: I just tried it and I got the following:

```
Error: Maximum update depth exceeded. This can happen when a component repeatedly calls setState inside componentWillUpdate or componentDidUpdate. 

React limits the number of nested updates to prevent infinite loops.
```

This probably happened because calling `this.computeSizes` this early doesn't work. The media must be loaded before this function is called. It calculated a zoom level of `NaN` and everything broke from there.

I also tried to remove the "this.computeSizes" callback and this cleared the error but the `initialCroppedAreaPixels` didn't work. I was back to the previous issue.


I don't have time to troubleshoot further.

ValentinH: Alright, let's call this good enough then :)

PR #553, opened 2024-08-08
manualResize added

alireza5014:
manualResize added


ValentinH:
Thanks for this contribution but this feature doesn't match the vision of this library.

PR #575, opened 2024-11-27
Updated keyboard support

IanSymplectic:
This PR updates keyboard support for moving the crop area in react-easy-crop using the arrow keys, addressing issue https://github.com/ValentinH/react-easy-crop/issues/574.

- Users can now hold the shift key when moving the crop area using the arrow keys, in order to move it by a smaller amount.
- Moving the crop area using the arrow keys no longer causes unintended scrolling.


ValentinH (approved):

ValentinH:
Thank you for this PR!

PR #583, opened 2025-02-12
🐛fix: make the prop CropperProps.keyboardStep not required

kvsur:
image

Prop of `CropperProps.keyboardStep`  is not required in the doc, change it into optional.

### DIFF
--- a/src/Cropper.tsx
+++ b/src/Cropper.tsx
@@ -62,7 +62,7 @@ export type CropperProps = {
   setMediaSize?: (size: MediaSize) => void
   setCropSize?: (size: Size) => void
   nonce?: string
-  keyboardStep: number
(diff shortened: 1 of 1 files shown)


src/Cropper.tsx
ValentinH, line 1: Actually, what issue are you trying to fix?
I understand that it's a bit weird but as long as the props are defined in the `defaultProps` static object, they are actually considered optional for the parent caller.

kvsur: Sorry, I configured tsconfig.json in a wrong way, which made the keyboardStep property required. I will withdraw this pull request

ValentinH: No worries!

PR #589, opened 2025-02-24
feat: emit crop data when a keyboard interaction is complete

### LINKED ISSUE #586: Emit crop data when keyboard interaction is complete
**Is your feature request related to a problem? Please describe.**
When a drag interaction is complete, the crop area is emitted and `onInteractionEnd` is called ([source](https://github.com/ValentinH/react-easy-crop/blob/d95e0ccb45f0210169fd260bb8d3a5db6c8d46e6/src/Cropper.tsx#L562-L563)). However, this does not happen for keyboard interactions, which only trigger [`onCropChange`](https://github.com/ValentinH/react-easy-crop/blob/d95e0ccb45f0210169fd260bb8d3a5db6c8d46e6/src/Cropper.tsx#L771) when a key is pressed (or held).

An issue can occur when users of react-easy-crop use `onCropComplete` to perform an action (like setting state). `onCropComplete` is potentially never called if the last interaction was a keydown.

**Describe the solution you'd like**
Seems like it would be good to introduce an event handler for `keyup` that emits the crop area and calls `onInteractionEnd` like `onDragStopped`.

priceld:
Fixes: #586 

This change also adds a call to `onInteractionStart` when a keyboard interaction starts (i.e. `keydown`) and a corresponding `onInteractionEnd` when it finishes (i.e. `keyup`). Those callbacks are only called when the key is an arrow key.

I did test this out with the demo app. Here is the console output when doing a single press of an arrow key:
```
user interaction started
onCropAreaChange! <callback data omitted>
onCropComplete! <callback data omitted>
onCropAreaChange! <callback data omitted>
user interaction ended
```
And then when pressing and holding an arrow key:
```
user interaction started
onCropAreaChange! <callback data omitted>
onCropAreaChange! <callback data omitted>
onCropAreaChange! <callback data omitted>
...
onCropComplete! <callback data omitted>
onCropAreaChange! <callback data omitted>
user interaction ended
```

> [!NOTE]
I did not add any e2e tests (there don't appear to be any for keyboard events anyway), but I can look into it if that is desirable.


src/Cropper.tsx
```diff
@@ -768,9 +768,28 @@ class Cropper extends React.Component<CropperProps, State> {
       newCrop = restrictPosition(newCrop, this.mediaSize, this.state.cropSize, zoom, rotation)
     }
 
+    if (!event.repeat) {
```
priceld, line 771: note: MDN did [mention](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/repeat#browser_compatibility) that Chrome and Opera have an [open bug](https://crbug.com/40940886) with around the `repeat` property.

If I'm understanding it correctly, it could mean that users on Windows and Linux could see multiple `onInteractionStart` calls when holding a key. I'm not sure if that is a big deal or not so I wanted to call it out.

ValentinH (approved):

ValentinH (approved):

ValentinH:
Thank you for taking the time to create this PR. This will automatically be released shortly. 🚀

ValentinH:
@allcontributors add @priceld for code

PR #593, opened 2025-02-25
feat: add cropperProps option to specify additional props for cropper

priceld:
This change adds a new  `cropperProps` option that allows users to specify additional props for the cropper element. I modeled this after the existing `mediaProps` option.

My specific use case for this is that I need to be able to specify an `aria-label` for the cropper since it can receive focus. This will allow for a more accessible workflow for users using screen readers.


ValentinH (approved):

ValentinH:
Thank you!

felipedeboni:
@ValentinH I was updating the lib and notice TS started complaining about it.

I think both `mediaProps` and `cropperProps` should be an optional params, as it avoids `cropperProps={{}}`.

Thoughts about it?

ValentinH:
I still don't fully understand how to type optional props for class components 😭

What we did so far was to not mark them as optional as long as we have corresponding default props defined, it was working properly.

Are you using the CropperProps or having the error directly on the component?

PR #605, opened 2025-05-21
fix: clean up gesturechange instead of gesturemove to re-allow zooming on iOS devices

mclavin:
There seems to be an issue with the the cropper not cleaning up 'gesturechange' properly when unmounting. This makes it almost impossible to zoom in the browser after interacting with the cropper area. 

This change removes 'gesturechange' instead of 'gesturemove'


ValentinH:
What a nice catch!
Could you also rename the `onGestureMove` method to `onGestureChange` for consistency please?

mclavin:
@ValentinH done :)

ValentinH (approved):

ValentinH:
Thank you for this fix!

ValentinH:
@allcontributors add @mclavin for code and bug fix

PR #622, opened 2025-09-15
fix: inconsistent crop area position after window resize

### LINKED ISSUE #621: Inconsistent crop area when resizing window (inside modal with no animation)
**Describe the bug**
If I place the crop area inside a modal (with no fading or scaling effect), drag the image to the left or right edge, and then resize the window, shrinking, then expanding, you will notice the crop area is slowly restored to the initial position. Is this a known issue? Can this be fixed in this lib? Let me know if anyone has had this issue before. I might take a look into this.

**To Reproduce**
1. Go to https://cjcg52.csb.app/
2. Click on 'Open editor'
3. Move the image all the way to the left
4. Decrease the window width to be similar to mobile, then increase to the initial position
5. You will notice the crop area is not restored to the initial version (centered).

**Expected behavior**
I think we should keep the correct image position regardless of the window size.

https://codesandbox.io/p/devbox/crimson-night-cjcg52?from-embed=

osnynetto:
This PR makes the crop selection layout-independent, so it stays exactly where the user left it when the window or modal size changes.

Fixes https://github.com/ValentinH/react-easy-crop/issues/621

### DIFF
--- a/src/Cropper.tsx
+++ b/src/Cropper.tsx
@@ -126,6 +126,8 @@ class Cropper extends React.Component<CropperProps, State> {
   currentDoc: Document | null = typeof document !== 'undefined' ? document : null
   currentWindow: Window | null = typeof window !== 'undefined' ? window : null
   resizeObserver: ResizeObserver | null = null
+  canonicalCroppedAreaPercentages: Area | null = null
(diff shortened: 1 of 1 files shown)


ValentinH:
Lovely, thank you very much!

PR #623, opened 2025-09-24
hotfix: Fix duplicate state updates

osnynetto:
This PR removes an unintended double setState sequence introduced in #622 that could trigger duplicate renders and callbacks (onCropChange, onZoomChange, onCropComplete) whenever cropSize changes.


ValentinH:
Thank you for taking care of this.

PR #630, opened 2025-11-14
fix: ESM build

### LINKED ISSUE #490: Cannot use import statement outside a module

anomiex:
The ESM build is not usable in strict tools because the file is named `index.module.js` and the package.json indicates that `.js` files are CommonJS rather than ESM.

It seems a pretty simple change to the build configuration to have it be named `index.module.mjs` instead.

Fixes #490

### DIFF
--- a/bili.config.ts
+++ b/bili.config.ts
@@ -23,7 +23,7 @@ const config: Config = {
       config.output.fileName = 'umd/react-easy-crop[min].js'
     }
     if (format === 'esm') {
-      config.output.fileName = '[name].module.js'
+      config.output.fileName = '[name].module.mjs'
     }
     return config
   },
--- a/scripts/copy-build-files.js
+++ b/scripts/copy-build-files.js
@@ -21,8 +21,8 @@ async function createPackageFile() {
     unpkg: './umd/react-easy-crop.js',
     jsdelivr: './umd/react-easy-crop.js',
     module: './index.module.mjs',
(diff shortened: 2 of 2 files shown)


ValentinH:
Thanks!

ValentinH:
@all-contributors add @anomiex for infrastructure

PR #653, opened 2026-07-02
fix: debounce onCropComplete/onCropAreaChange during resize bursts

### LINKED ISSUE #309: onCropComplete is firing on window resize
**Describe the bug**
If you resize the window, onCropComplete fires.

**To Reproduce**
1. Go to [this codesandbox](https://codesandbox.io/s/react-easy-crop-forked-p9r34?file=/src/index.js)
2. Put a console log in the the onCropComplete
3. Resize the window
4. Notice console logs with what seems like with no debounce

**Expected behavior**
Not expecting onCropComplete to fire on window resize.. i understand why this might be in place for responsive reasons. But if it is a required feature I would expect the window resize to be debounced.

HyeryongChoi:
## Summary

Fixes #309.

`recomputeCropPosition` always calls `emitCropData()` unconditionally, and `computeSizes` now runs on every `ResizeObserver` tick. As a result, dragging the window/container edge fires `onCropComplete` and `onCropAreaChange` dozens of times during a single resize, even though the crop data at each intermediate tick isn't meaningful to consumers.

This PR debounces the `emitCropData()` call specifically for resize-triggered `computeSizes` calls (250ms), while every other trigger (rotation/aspect/zoom/cropSize prop changes, media load, drag/wheel/pinch interactions) keeps firing `onCropComplete` immediately as before. The crop area still updates visually in real time while resizing — only the completion callbacks are debounced.

## Changes

- `computeSizes(isResizeTriggered)` and `recomputeCropPosition(isResizeTriggered)` now take a flag, set to `true` only when called from the `ResizeObserver` callback or the legacy `window resize` fallback listener.
- Added `debouncedEmitCropData()`, mirroring the existing `wheelTimer` debounce pattern already used for `onInteractionStart`/`onInteractionEnd` on wheel zoom.
- Cleared the new timer in `componentWillUnmount` to avoid calling back into an unmounted component.

## Test plan

- [x] `pnpm unit` — 38 passing
- [x] `pnpm type-check` — clean
- [x] `pnpm lint` — clean
- [x] `pnpm build` — succeeds
- [x] Added a new Cypress e2e test (`should debounce onCropComplete during a burst of window resizes`) that triggers a burst of real resizes and asserts `onCropComplete` fires exactly once after the debounce window, instead of once per resize tick.
- [x] Verified the new test fails against the pre-fix code (reverted the source change locally, confirmed the test catches the regression) and passes with the fix.
- [x] Ran `pnpm start` and manually resized the demo window — crop area still tracks smoothly during the resize.

### DIFF
--- a/src/Cropper.tsx
+++ b/src/Cropper.tsx
@@ -15,6 +15,11 @@ import {
 } from './helpers'
 import cssStyles from './styles.css?raw'
 
+// how long to wait after the last resize-triggered size recompute before
+// firing onCropComplete/onCropAreaChange, so a window/container resize
+// doesn't spam consumers with callbacks on every frame
+const RESIZE_EMIT_DEBOUNCE_TIME = 250
+
 export type CropperProps = {
   image?: string
   video?: string | VideoSrc[]
@@ -123,6 +128,7 @@ class Cropper extends React.Component<CropperProps, State> {
   rafDragTimeout: number | null = null
   rafPinchTimeout: number | null = null
   wheelTimer: number | null = null
+  resizeEmitTimer: number | null = null
   currentDoc: Document | null = typeof document !== 'undefined' ? document : null
   currentWindow: Window | null = typeof window !== 'undefined' ? window : null
   resizeObserver: ResizeObserver | null = null
@@ -148,7 +154,7 @@ class Cropper extends React.Component<CropperProps, State> {
       this.initResizeObserver()
       // only add window resize listener if ResizeObserver is not supported. Otherwise, it would be redundant
       if (typeof window.ResizeObserver === 'undefined') {
-        this.currentWindow.addEventListener('resize', this.computeSizes)
+        this.currentWindow.addEventListener('resize', this.onWindowResize)
       }
       this.props.zoomWithScroll &&
         this.containerRef.addEventListener('wheel', this.onWheel, { passive: false })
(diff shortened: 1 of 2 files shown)


HyeryongChoi:
Thanks for the review, Copilot.

- **Stale duplicate emit after a resize + interaction**: valid, fixed in 5b5ad48 — `emitCropData()` now clears any pending resize debounce timer before emitting, so an immediate emit (drag stop, keyboard end, etc.) supersedes a stale resize-triggered one instead of both firing.
- **Not nulling `resizeEmitTimer` after clearing in `componentWillUnmount`**: left as-is to match the existing cleanup pattern used elsewhere in this file for `wheelTimer` and the `resizeObserver` ref, which also aren't nulled out on unmount.

src/Cropper.tsx
```diff
@@ -348,7 +361,7 @@ class Cropper extends React.Component<CropperProps, State> {
     return this.props.objectFit
   }
 
-  computeSizes = () => {
+  computeSizes = (isResizeTriggered = false) => {
```
ValentinH: nitpick: let's used a destructured object to have a clear name for the parameter in the caller.

```suggestion
  computeSizes = ({ isResizeTriggered } = { isResizeTriggered = false }) => {
```

src/Cropper.tsx
```diff
@@ -728,7 +746,7 @@ class Cropper extends React.Component<CropperProps, State> {
     }
   }
 
-  recomputeCropPosition = () => {
+  recomputeCropPosition = (isResizeTriggered = false) => {
```
ValentinH: same here

src/Cropper.tsx
HyeryongChoi: Done in 3f00504 — switched to a destructured `{ isResizeTriggered }` param on both `computeSizes` and `recomputeCropPosition`, typed as `{ isResizeTriggered = false }: { isResizeTriggered?: boolean } = {}` to satisfy TypeScript.

src/Cropper.tsx
HyeryongChoi: Same fix applied here too, in 3f00504.

ValentinH (approved):

ValentinH:
Thank you!

PR #664, opened 2026-08-11
Migrate Cropper to Hooks and React Compiler

ValentinH:
## Summary

- Replaces the class component with a Hooks engine using one mount effect plus an internal SSR-safe layout effect.
- Targets React 19.2 and React Compiler; no manual callback or memo hooks.
- Modernizes refs, stylesheet handling, callback semantics, demo, docs, and migration guidance.
- Adds StrictMode, SSR, interaction lifecycle, media-switch, and built-package CJS/ESM coverage.

## Breaking changes

- Requires React 19.2.
- Component ref now targets the outer DOM container; legacy setter refs and the public class instance are removed.
- Automatic CSS uses React-managed stylesheet resources, with an inline nonce path for SSR.
- Callback emissions are deduplicated and tied to committed controlled values.

## Verification

- pnpm type-check
- pnpm lint
- pnpm unit (54 tests)
- pnpm build
- pnpm docs:build
- pnpm exec vite build ./examples --config ./vite.config.mts
- pnpm e2e (28 tests)

Release: major.

<details>
  <summary>📦 Published PR as canary version: <code>7.0.0--canary.664.2c66a6d.0</code></summary>
  <br />

  :sparkles: Test out this PR locally via:
  
  ```bash
  npm install react-easy-crop@7.0.0--canary.664.2c66a6d.0
  # or 
  yarn add react-easy-crop@7.0.0--canary.664.2c66a6d.0
  ```
</details>


ValentinH (commented):
## Standards

No documented standards; no hard violations. All findings are baseline judgement calls:

- **Repeated Switches** — `src/useCropper.tsx:1184-1228`: both handlers repeat `switch (event.key)` with `case 'ArrowUp'`, `'ArrowDown'`, `'ArrowLeft'`, and `'ArrowRight'`. Define one arrow-delta map/type guard and reuse it.

- **Duplicated Code** — `src/useCropper.tsx:1116-1144`: `function setImageElement(...)` and `function setVideoElement(...)` repeat `runtime.media = element`, `const refCleanup = assignRef(mediaRef, element)`, and equivalent cleanup. Extract a typed media-ref callback factory/helper.

- **Data Clumps** — `src/useCropper.tsx:102-103,358-359,1052-1061`: `cropSizeWidth?: number` and `cropSizeHeight?: number` always travel and compare together, including `previous.cropSizeWidth !== requestedCropSize?.width` and the matching height check. Store `cropSize?: Size` in `CommittedInputs` and compare through `sizesEqual`; scalar effect dependencies may remain separate.

- **Duplicated Code** — `tsdown.config.mts:29-32`, `vite.config.mts:8-11`, `vitest.config.mts:7-11`: all repeat `presets: [reactCompilerPreset({ target: '19', panicThreshold: 'all_errors' })]`. Centralize the compiler options/preset factory so build, demo, and tests cannot drift.

## Spec

- **Incorrect implementation** — “Use all the best practices” / “Rely on the React compiler…”: `src/useCropper.tsx:616-903` creates Effect Events subsequently passed to `requestAnimationFrame`, timers, and document listeners from ordinary interaction paths entered via JSX handlers at `src/useCropper.tsx:1155-1171`. [React permits Effect Events only from Effects or other Effect Events](https://react.dev/reference/react/useEffectEvent#caveats); the `React.useEffectEvent` namespace spelling currently bypasses the hooks linter's call-site check. Route these async callbacks through effect-owned subscriptions (or a non-Effect-Event latest-handler primitive) and reserve Effect Events for effect-owned calls.

- **Partial migration** — “Today is the day we finally migrate to Hooks.”: `docs/package.json:22` still installs `react-easy-crop` 5.5.6. The deployed documentation examples import that package, so they continue running the old class implementation while the docs describe v7 behavior; the new docs CI build validates neither the hooks engine nor its stylesheet/callback changes. Make the docs consume the local v7 build/workspace package.

Summary: Standards — 4 judgement-call findings; worst: duplicated Compiler configuration can drift across build, demo, and tests. Spec — 2 findings; worst: interaction-path Effect Events violate React's call-site contract.

ValentinH:
Closing this PR after reviewing the trade-offs. The migration increases source complexity and bundle size, introduces breaking API changes, and requires React 19.2, without a demonstrated consumer benefit sufficient to justify those costs.

The existing class implementation remains supported by React. For this library, keeping the implementation small and stable takes priority over adopting newer patterns.

There is useful work here, particularly the regression tests and SSR/callback improvements, which we can evaluate separately. We can revisit a hooks migration if it enables a concrete improvement for consumers.

